PR #1, opened 2018-12-13
Automate collector deployment and teardown

bdunne:
Handles:
- newly added sources
- deleted sources
- changes to sources, endpoints and authentications

Depends on https://github.com/ManageIQ/topological_inventory-api/pull/74


lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -116,9 +97,67 @@ def pg_connection_args
...
+        return unless digest
+        deployment = object_manager.get_deployment_configs("topological-inventory/collector=true").detect { |i| i.metadata.labels["topological-inventory/collector_digest"] == digest }
+        return unless deployment
+        puts "Removing objects for deployment #{deployment.metadata.name}"
+        object_manager.delete_deployment_config(deployment.metadata.name)
+        object_manager.delete_secret("#{deployment.metadata.name}-secrets")
+      end
+
+      def collector_deployment_name_for_source(source)
+        "topological-inventory-collector-source-#{source["source_id"]}"
```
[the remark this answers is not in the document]
carbonin: Maybe we should use the digest here? Even though that's a bit gross ... :worried:

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,97 +1,77 @@
...
-          source_for_deployment(deployment) == source["uid"]
-        end
-      end
-
-      def create_objects_for_source(source)
-        @object_manager.create_deployment_config(deployment_name_for_source(source)) do |d|
-          container = d[:spec][:template][:spec][:containers].first
-          container[:env] << collector_container_environment(source)
-          container[:image] = "#{ENV["MY_NAMESPACE"]}/topological-inventory-collector-openshift:latest"
+          sleep 10
```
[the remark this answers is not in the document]
bdunne: Do you have an example use case?

lib/topological_inventory/orchestrator/worker.rb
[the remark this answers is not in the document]
bdunne: Let's cross that bridge when we get to it.  For now, since sources to collectors is 1 to 1, this helps identify which source this collector is for.

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -116,9 +97,67 @@ def pg_connection_args
...
+
+
+      ### Openshift stuff
+      def collectors_from_openshift
+        object_manager.get_deployment_configs("topological-inventory/collector=true").collect { |i| i.metadata.labels["topological-inventory/collector_digest"] }
+      end
+
+      def create_openshift_objects_for_source(digest, source)
+        puts "Creating objects for source #{source["source_id"]} with digest #{digest}"
+        object_manager.create_secret(collector_deployment_secret_name_for_source(source), {"openshift-token" => source["token"]})
```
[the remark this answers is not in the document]
bdunne: @agrare Any thoughts?

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -116,9 +97,67 @@ def pg_connection_args
...
+        object_manager.create_deployment_config(collector_deployment_name_for_source(source)) do |d|
+          d[:metadata][:labels]["topological-inventory/collector_digest"] = digest
+          d[:metadata][:labels]["topological-inventory/collector"] = "true"
```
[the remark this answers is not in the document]
bdunne: I like how clean the `ObjectManager` definitions are and would prefer not to put collector deployment specifics in there.

lib/topological_inventory/orchestrator/worker.rb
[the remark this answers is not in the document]
bdunne: I don't really like `id` because it could be confused with the `id` of the record.  Maybe `public` and `private`?

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,97 +1,77 @@
...
+              api_get("sources/#{source["id"]}/endpoints").each do |endpoint|
+                definition_information = collector_definition[endpoint["role"]]
+                next unless definition_information
+                array << database.add(
+                  {
+                    "host"       => endpoint["host"],
+                    "image"      => definition_information["image"],
+                    "source_id"  => source["id"],
+                    "source_uid" => source["uid"],
+                    "token"      => authentication_for_endpoint(endpoint["id"].to_i)["password"],
```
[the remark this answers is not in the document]
bdunne: I'm thinking about changing this key to `secret` and the value to a hash.  Let's discuss in the comment below

lib/topological_inventory/orchestrator/worker.rb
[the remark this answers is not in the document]
agrare: Hey sorry didn't see this, providers fall into three groups:
1. Typical username / password (most infra providers)
2. public / private keys (aws/azure)
3. Just a private key (openshift/kubernetes)

Public/private would cover all these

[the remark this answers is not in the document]
bdunne: After some more discussions with @carbonin and @Fryguy we decided `private` and `public` were probably not great names since they're both ruby keywords.  However `username` and `password` are the very popular and clearly understood, so we'll just stick with that for now.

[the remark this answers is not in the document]
agrare: :+1: awesome

bdunne:
Okay @carbonin I was able to get a test working.  Let me know if you think anything else needs tests.

Gemfile
```diff
@@ -1,11 +1,3 @@
...
-# For MiqPassword
-gem "manageiq-gems-pending", :git => "https://github.com/ManageIQ/manageiq-gems-pending.git", :branch => "master"
-
-# For starting collector pods
-gem "kubeclient",  "~>4.0"
-# For reading from Authentications
-gem "pg",          "~>1.1"
-# For reading from Sources + Endpoints
-gem "rest-client", "~>2.0"
+gemspec
```
carbonin: Why make this a gem?

bin/topological_inventory-orchestrator
```diff
@@ -1,9 +1,12 @@
 #!/usr/bin/env ruby
 
+Signal.trap("TERM") { exit }
```
carbonin: :+1:

config/collector_definitions.yaml
```diff
@@ -21,12 +21,12 @@
 ---
 openshift:
   default:
-    image: "topological-inventory-ci/topological-inventory-collector-openshift:latest"
+    image: "docker-registry.default.svc:5000/topological-inventory-ci/topological-inventory-collector-openshift:latest"
```
carbonin: I'm not sure why this change was needed. I think this should probably be removed before we merge.

Also, I'm not sure where we will actually get the image from. I doubt it will be in the `topological-inventory-ci` namespace. Should we publish those on dockerhub or something?

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,97 +1,79 @@
...
+        puts "Checking..."
+
+        (current - expected).each { |i| remove_openshift_objects_for_source(i) }
+        (expected - current).each { |i| create_openshift_objects_for_source(i, database[i]) }
+
+        puts "Checking... complete."
+      end
+
+
+      ### API STUFF
```
carbonin: Follow up, but instead of these comments should these be separate objects? Maybe we could compose these functions better.

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -102,12 +84,13 @@ def authentication_for_endpoint(endpoint_id)
             resource_id = $1
         SQL
         conn.exec_params(sql, [endpoint_id]).first.tap do |auth|
+          next if auth.nil?
           auth["password"] = MiqPassword.decrypt(auth["password"])
```
carbonin: Doesn't this have to change to use `ManageIQ::Password`?

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -116,9 +99,64 @@ def pg_connection_args
...
+      def collector_deployment_secret_name_for_source(source)
+        "#{collector_deployment_name_for_source(source)}-secrets"
+      end
+
+      def collector_container_environment(source)
+        secret_name = "#{collector_deployment_name_for_source(source)}-secrets"
+        [
+          {:name => "AUTH_PASSWORD", :valueFrom => {:secretKeyRef => {:name => secret_name, :key => "password"}}},
+          {:name => "AUTH_USERNAME", :valueFrom => {:secretKeyRef => {:name => secret_name, :key => "username"}}},
+          {:name => "INGRESS_API", :value => ENV["TOPOLOGICAL_INVENTORY_INGRESS_API_PORT"].sub("tcp://", "http://")},
```
carbonin: I think I'd prefer this be composed as:
```ruby
"http://#{ENV["TOPOLOGICAL_INVENTORY_INGRESS_API_SERVICE_HOST"]}:#{ENV["TOPOLOGICAL_INVENTORY_INGRESS_API_SERVICE_PORT"]}"
```

Alternatively we could pass it into the orchestrator as a single ENV variable like we are doing currently for collectors: https://github.com/ManageIQ/topological_inventory-deploy/blob/e8d28a52e30fac40ad03c9ee3ea74abb4cb8f1e1/amazon-collector/deploy_template.yaml#L43

Gemfile
bdunne: It's much easier to get rspec and everything else set up.  We don't need to release it to rubygems.

config/collector_definitions.yaml
bdunne: The old address didn't work for me in the test environment.

> Also, I'm not sure where we will actually get the image from. I doubt it will be in the topological-inventory-ci namespace. Should we publish those on dockerhub or something?

I like the idea of it being published somewhere, but I think we won't be pulling from dockerhub in production.

carbonin: Agreed, I'm just having trouble coming up with a sane default value.

bdunne:
@carbonin Any other concerns?

carbonin:
I think the only thing I still want to discuss is the gem structure.

I think we could just move the gemspec deps into the Gemfile and drop the version file entirely, right?
I feel like this looking like an application rather than a library is helpful in understanding its purpose.

Gemfile
Fryguy: If it's an app it should not be a gem.  It's confusing.

lib/topological_inventory/orchestrator/object_database.rb
```diff
@@ -0,0 +1,24 @@
+module TopologicalInventory
+  module Orchestrator
+    class ObjectDatabase
```
Fryguy: This whole thing feels overkill to me, as what you have built is basically a specialized Hash, where the key is a one-line transformation.  That is, instead of db.add earlier, I could just see `db = {}; db[key_from_object(object)] = object` (more or less)

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,97 +1,79 @@
...
-        collector_deployments.each do |deployment|
-          unless sources.include?(source_for_deployment(deployment))
-            @object_manager.delete_deployment_config(deployment.metadata.name)
-            @object_manager.delete_secret("#{deployment.metadata.name}-secrets")
+      def make_openshift_match_database
+        database = ObjectDatabase.new
+        expected = collectors_from_database(database)
+        current  = collectors_from_openshift
+
+        puts "Checking..."
```
Fryguy: puts vs log?

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,97 +1,79 @@
...
+
+
+      ### API STUFF
+      def collectors_from_database(database)
+        [].tap do |array|
+          api_get("source_types").each do |source_type|
+            collector_definition = collector_definitions[source_type["name"]]
+            next if collector_definition.nil?
+            api_get("source_types/#{source_type["id"]}/sources").each do |source|
+              api_get("sources/#{source["id"]}/endpoints").each do |endpoint|
```
Fryguy: code-wise this is hard to read, and even if it switched to use a generated client I'm not sure it would get much better...I think splitting the nested looping from the "work" would help...something like

```ruby
def each_collector_from_database(database)
  api_get("source_types").each do |source_type|
    collector_definition = collector_definitions[source_type["name"]]
    next if collector_definition.nil?
    api_get("source_types/#{source_type["id"]}/sources").each do |source|
      api_get("sources/#{source["id"]}/endpoints").each do |endpoint|
        definition_information = collector_definition[endpoint["role"]]
        next unless definition_information
        yield source, definition_information, endpoint
      end
    end
  end
end

def collectors_from_database(database)
  [].tap do |array|
    each_collector_from_database do |source, definition_information, endpoint|
      auth = authentication_for_endpoint(endpoint["id"].to_i)
      array << database.add(
        {
          "host"       => endpoint["host"],
          "image"      => definition_information["image"],
          "source_id"  => source["id"],
          "source_uid" => source["uid"],
          "secret"     => {
            "password" => auth["password"],
            "username" => auth["username"],
          },
        }
      )
    end
  end
end
```

Then, once you do that, I think the `[].tap` can more naturally turn into a map operation, and thus won't look as strange.

bdunne:
Marking as WIP for now because I'm trying to get pagination done first and it will require some changes to the api response parsing.  Feel free to review the rest of the changes though.

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,97 +1,100 @@
...
-      def secret_name_for_source(source)
-        "#{deployment_name_for_source(source)}-secrets"
-      end
 
-      def deployment_name_for_source(source)
-        "topological-inventory-collector-source-#{source["id"]}"
+      ### API STUFF
+      def each_endpoint
+        return enum_for(:each_endpoint) unless block_given?
+        each_resource(url_for("source_types")) do |source_type|
```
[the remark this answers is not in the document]
bdunne: We can't move the `url_for` calls into `each_resource` since it is also called with full URL reference to the next page.

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -116,9 +120,64 @@ def pg_connection_args
...
+        "#{collector_deployment_name_for_source(source)}-secrets"
+      end
+
+      def collector_container_environment(source)
+        secret_name = "#{collector_deployment_name_for_source(source)}-secrets"
+        [
+          {:name => "AUTH_PASSWORD", :valueFrom => {:secretKeyRef => {:name => secret_name, :key => "password"}}},
+          {:name => "AUTH_USERNAME", :valueFrom => {:secretKeyRef => {:name => secret_name, :key => "username"}}},
+          {:name => "INGRESS_API", :value => "http://#{ENV["TOPOLOGICAL_INVENTORY_INGRESS_API_SERVICE_HOST"]}:#{ENV["TOPOLOGICAL_INVENTORY_INGRESS_API_SERVICE_PORT"]}"},
+          {:name => "OPENSHIFT_HOSTNAME", :value => source["host"]},
```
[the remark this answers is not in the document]
bdunne: yep, I'll fix that

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,97 +1,100 @@
...
 
-      def secret_name_for_source(source)
-        "#{deployment_name_for_source(source)}-secrets"
-      end
 
-      def deployment_name_for_source(source)
-        "topological-inventory-collector-source-#{source["id"]}"
+      ### API STUFF
+      def each_endpoint
+        return enum_for(:each_endpoint) unless block_given?
```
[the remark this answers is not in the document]
carbonin: Oh I misread the caller. The block is being sent to collect, not `each_endpoint`.

That said, this still seems harder to read to me. I'm still not really sure how this works after reading the `enum_for` documentation...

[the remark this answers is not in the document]
bdunne: `enum_for` returns an Enumerator object that will pass the yielded values to the caller.

lib/topological_inventory/orchestrator/worker.rb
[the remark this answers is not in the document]
carbonin: Ah, okay, and that link will already have the full url, got it.

lib/topological_inventory/orchestrator/worker.rb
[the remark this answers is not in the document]
carbonin: Okay, but I still don't see why this method has to account for being called with a block if we are not going to do that. Were dealing with two cases when we only have one.

Simply, why not make this method return an array rather than yielding?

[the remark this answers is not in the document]
Fryguy: I think what @carbonin is saying is, why do the enumerator dance when this method can just return an Array (i.e. name the method `endpoints` instead of `each_endpoint`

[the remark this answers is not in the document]
bdunne: Oh, I think the array is a bad idea as far as memory is concerned.  I'd rather only keep the information that we want for as long as it is needed.  This way we can process each resource as we fetch them from the API rather than fetching everything up front then looping over the data.

lib/topological_inventory/orchestrator/worker.rb
bdunne: Done

.gitignore
```diff
@@ -1 +1,2 @@
 Gemfile.lock
+v2_key
```
carbonin: Do we still expect this to be here?

Gemfile
```diff
@@ -1,11 +1,12 @@
...
-# For MiqPassword
-gem "manageiq-gems-pending", :git => "https://github.com/ManageIQ/manageiq-gems-pending.git", :branch => "master"
-
-# For starting collector pods
-gem "kubeclient",  "~>4.0"
-# For reading from Authentications
-gem "pg",          "~>1.1"
-# For reading from Sources + Endpoints
+gem "manageiq-loggers", "~>0.1"
+gem "manageiq-password", "~>0.2"
```
carbonin: Do we still need this?

config/collector_definitions.yaml
```diff
@@ -21,12 +21,12 @@
 ---
 openshift:
   default:
```
carbonin: Lets remove the endpoint role and `sources_per_deployment` concept for now.

We're not really using them and don't have a clear usecase for either until we get new source types.
At that point we'll probably have to refactor the collectors to read auth from a configmap anyway so adding this back in will be the least of our concerns.

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,124 +1,168 @@
...
-      attr_reader :api_url, :ingress_url
+      attr_reader :logger
 
-      def initialize(api_url, ingress_url, collector_definitions_file)
-        raise "Access to the cluster using serviceaccounts is required" unless ObjectManager.available?
-        @api_url               = api_url
-        @ingress_url           = ingress_url
-        @object_manager        = ObjectManager.new
-        @collector_definitions = YAML.load_file(collector_definitions_file)
+      def initialize(api_base_url: ENV["API_URL"], collector_definitions_file: ENV["CDF"])
```
carbonin: Can we make the ENV var non-abbreviated?

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,124 +1,168 @@
...
+
+
+      ### API STUFF
+      def each_endpoint
+        return enum_for(:each_endpoint) unless block_given?
+        each_resource(url_for("source_types")) do |source_type|
+          collector_definition = collector_definitions[source_type["name"]]
+          next if collector_definition.nil?
+          each_resource(url_for("source_types/#{source_type["id"]}/sources")) do |source|
+            each_resource(url_for("sources/#{source["id"]}/endpoints")) do |endpoint|
```
carbonin: If we assume one endpoint per source and one auth per endpoint can we dump two levels of this loop?

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,124 +1,168 @@
...
+              each_resource(url_for("authentications?resource_type=Endpoint&resource_id=#{endpoint["id"]}&authtype=default")) do |authentication|
```
carbonin: What are the possible values for `authtype`? Right now we have it as `nil`, but will it always be `default` or should we just take the first one?

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,124 +1,168 @@
...
-        @sources = nil
-        @collector_deployments = nil
+      def collector_container_environment(source)
+        secret_name = "#{collector_deployment_name_for_source(source)}-secrets"
+        [
+          {:name => "AUTH_PASSWORD", :valueFrom => {:secretKeyRef => {:name => secret_name, :key => "password"}}},
+          {:name => "AUTH_USERNAME", :valueFrom => {:secretKeyRef => {:name => secret_name, :key => "username"}}},
+          {:name => "INGRESS_API", :value => "http://#{ENV["TOPOLOGICAL_INVENTORY_INGRESS_API_SERVICE_HOST"]}:#{ENV["TOPOLOGICAL_INVENTORY_INGRESS_API_SERVICE_PORT"]}"},
+          {:name => "SOURCE_HOST", :value => source["host"]},
+          {:name => "SOURCE_UID",  :value => source["source_uid"]},
```
carbonin: Can we make this match https://github.com/ManageIQ/topological_inventory-collector-openshift/pull/34

agrare:
@carbonin @bdunne  https://github.com/ManageIQ/topological_inventory-collector-openshift/pull/34 and https://github.com/ManageIQ/topological_inventory-collector-amazon/pull/9

lib/topological_inventory/orchestrator/worker.rb
bdunne: @agrare ?

agrare: This is supposed to be if this was e.g. a username+password/ssh token/etc.. if providers support multiple types of creds.

Having a column that represents if the authentication is default for the endpoint would be useful though if we end up in a situation where we need to store multiple user credentials for the same endpoint for e.g. running operations as a user on a provider without impersonation.

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,124 +1,168 @@
...
+        each_resource(response["links"]["next"], &block)
       end
 
-      def secret_name_for_source(source)
-        "#{deployment_name_for_source(source)}-secrets"
+
+      # HACK for Authentications
+      def authentication_with_password(id)
+        url = URI.parse(@api_base_url).tap do |uri|
+          uri.path = "/internal/v0.0/authentications/#{id}"
```
carbonin: Is this the only version for the internal API?

bdunne: yes

lib/topological_inventory/orchestrator/worker.rb
bdunne: It sounds like we would need a validation around this to verify that there is only one default for the endpoint.

bdunne: >where we need to store multiple user credentials for the same endpoint

This may make it more difficult to transition away from the authentications table when the replacement service(s) become available

agrare: We would still have our own auth records for metadata just would fetch the password and maybe username from the vault service.

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -1,124 +1,173 @@
...
-      
-      def remove_collectors_for_deleted_sources
-        collector_deployments.each do |deployment|
-          unless sources.include?(source_for_deployment(deployment))
-            @object_manager.delete_deployment_config(deployment.metadata.name)
-            @object_manager.delete_secret("#{deployment.metadata.name}-secrets")
+
+
+      ### API STUFF
+      def each_endpoint
```
carbonin: Should this be `each_source` now?

carbonin:
Kicking travis as I think it's enabled now.

carbonin:
Ah there's no travis config yet. @Fryguy did mention he enabled it though, do you want to add one in this PR @bdunne ? Otherwise I can just merge and we can add it in a follow-up.

bdunne:
> Ah there's no travis config yet.

I'd rather do this in another PR

PR #3, opened 2019-02-28
Use the built-in environment variables for the API service

carbonin:
This allows us to just pass the same vars that we are already giving
to the other services (PATH_PREFIX and APP_NAME).

Additionally we encode the api version in the source rather than
passing it as a part of a path environment variable.

If we were to pass it in, we run the risk of the code expecting
one version, but the API endpoint providing a different one. This
way, we will always be in sync.

### DIFF
--- a/lib/topological_inventory/orchestrator/worker.rb
+++ b/lib/topological_inventory/orchestrator/worker.rb
@@ -11,7 +11,7 @@ module TopologicalInventory
     class Worker
       attr_reader :logger
 
-      def initialize(api_base_url: ENV["API_URL"], collector_definitions_file: ENV["CONTAINER_DEFINITIONS_FILE"])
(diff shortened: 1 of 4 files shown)


lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -27,6 +28,22 @@ def run
 
       private
 
+      def api_base_url
+        return @api_base_url if @api_base_url
```
bdunne: Prefer the `||=` syntax:
```ruby
@api_base_url ||= begin
  ...
end
```

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -27,6 +28,22 @@ def run
...
+
+        path = ENV["PATH_PREFIX"].to_s
+        path = "/#{path}" unless path.start_with?('/')
```
bdunne: :man_facepalming:

PR #4, opened 2019-03-04
Add the image trigger and alter image handling

carbonin:
This PR alters how the image for a collector is specified. Instead of listing out the fully qualified image, we now can specify the org/namespace of the image separately and default that to the `IMAGE_NAMESPACE` environment variable.

This is provided to most templates at deploy time and can be used here as well. The assumption is that the collector images will live in the same namespace as the orchestrator image. If that is not the case, the image namespace can be specified separately in the collector definitions file.

Additionally this PR fixes a small issue by calling `.to_s` on the port value in the collector deployment config.

### DIFF
--- a/config/collector_definitions.yaml
+++ b/config/collector_definitions.yaml
@@ -1,7 +1,8 @@
 # Layout of this file
 # <source_type>:
+#   image_namespace: <namespace> -- optional
 #   image: <image_ref>
 
 ---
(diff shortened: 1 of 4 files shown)


carbonin:
@bdunne @Fryguy I think this is the last patch I needed in this repo to get the orchestrator working in the lab cluster.

PR #5, opened 2019-03-04
Add the amazon collector to the definitions file

carbonin:


config/collector_definitions.yaml
```diff
@@ -6,3 +6,5 @@
 ---
 openshift:
   image: "topological-inventory-openshift:latest"
+amazon:
+  image: "topological-inventory-amazon-collector:latest"
```
bdunne: Will this eventually be renamed to match the openshift pattern?

carbonin: Probably, but it isn't as of yet. When we change it it should change the digest, so we will stop the old one and start a new one right?

bdunne: yes, just checking

PR #9, opened 2019-04-04
Deal with tenancy 😎

carbonin:
This PR makes the orchestrator pass a tenant header with each of its requests.

To query for sources we have to first use the internal API to query for tenants. For this call (and all others to the internal API) we use a special tenant identifier which won't collide with existing users. After we have the list of tenants we can then query for sources and their dependent objects tenant-by-tenant.

Also of note here, this PR changes the `#each_resource` method so that it can deal with both paginated and non-paginated responses as the internal API doesn't use paginated responses.

This will resolve https://projects.engineering.redhat.com/browse/TPINVTRY-282

### DIFF
--- a/lib/topological_inventory/orchestrator/worker.rb
+++ b/lib/topological_inventory/orchestrator/worker.rb
@@ -1,3 +1,4 @@
+require "base64"
 require "json"
 require "manageiq-loggers"
 require "manageiq-password"
@@ -10,6 +11,7 @@ module TopologicalInventory
   module Orchestrator
     class Worker
       API_VERSION = "v0.1".freeze
+      ORCHESTRATOR_TENANT = "system_orchestrator".freeze
 
       attr_reader :logger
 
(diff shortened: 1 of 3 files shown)


lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -107,8 +107,12 @@ def url_for(path)
...
         return if url.nil?
         response = get_and_parse(url)
-        response["data"].each { |i| yield i }
-        each_resource(response["links"]["next"], &block)
+        paging = response.is_a?(Hash)
+
+        resources = paging ? response["data"] : response
+        resources.each { |i| yield i }
+
+        each_resource(response["links"]["next"], &block) if paging
```
Fryguy: Minor but would fetch_path be safer here in case the links section doesn't exist?

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -100,6 +100,13 @@ def collectors_from_sources_api
         hash
       end
 
+      def internal_url_for(path, query = nil)
+        URI.parse(api_base_url).tap do |uri|
+          uri.path = File.join("/internal/v0.0/", path)
```
Fryguy: Ultimately we should do this kind of thing in a generic place in the lib directory.  There is an effort by, I think, @agrare to create a client wrapper in manageiq-api-common that can do auto-tenancy, but the general idea is you would set the version number at some common location like an initializer, so we don't have to bury it in the code itself.

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -63,39 +65,41 @@ def make_openshift_match_database
...
-            "source_uid"      => source["uid"],
-            "secret"          => {
-              "password" => auth["password"],
-              "username" => auth["username"],
-            },
-          }
-          key = digest(value)
-          hash[key] = value
+        each_tenant do |tenant|
+          each_source(tenant) do |source, endpoint, authentication, collector_definition|
```
Fryguy: Minor, but structurally something feels off about this...it reads funny to me that we would nest each_* in this `collectors_from_sources_api` method, as opposed to having that nesting live somewhere else, perhaps with a different name.  That is, I'd expect this method to call each_source, and we just refactor the guts of each_source to be tenant aware.

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -111,21 +115,31 @@ def url_for(path)
...
 
-      def get_and_parse(url)
-        JSON.parse(RestClient.get(url))
+      def get_and_parse(url, tenant_account = ORCHESTRATOR_TENANT)
+        JSON.parse(
+          RestClient.get(
+            url,
+            "x-rh-identity" => Base64.encode64(
+              {"identity" => {"account_number" => tenant_account}}.to_json
+            ).strip
```
Fryguy: This feels like something we shouldn't do here, but I'm ok temporarily.  I'd expect a) building identity would live in common, so we are not duplicating this common thing and b) we wouldn't use RestClient, but a proper client library.

Fryguy:
Overall looks good to me...some structural questions, but I'd be ok with merging this now and cleaning up later.

lib/topological_inventory/orchestrator/worker.rb
carbonin: Opened https://github.com/ManageIQ/topological_inventory-orchestrator/issues/11 for this

lib/topological_inventory/orchestrator/worker.rb
carbonin: Opened https://github.com/ManageIQ/topological_inventory-orchestrator/issues/10 for this

PR #15, opened 2019-04-11
Use the sources service for the orchestrator

agrare:
Update to use the sources service for getting source_types, sources,
endpoints, and authentications for starting collectors.

Depends on: https://github.com/RedHatInsights/e2e-deploy/pull/241

### DIFF
--- a/lib/topological_inventory/orchestrator/worker.rb
+++ b/lib/topological_inventory/orchestrator/worker.rb
@@ -11,18 +11,20 @@ require "topological_inventory/orchestrator/object_manager"
 module TopologicalInventory
   module Orchestrator
     class Worker
-      TOPOLOGY_API_VERSION = "v0.1".freeze
-      SOURCES_API_VERSION = "v1.0".freeze
       ORCHESTRATOR_TENANT = "system_orchestrator".freeze
 
-      attr_reader :logger, :sources_url, :topology_url
+      attr_reader :logger, :sources_api, :sources_internal_api, :topology_api, :topology_internal_api
 
-      def initialize(sources_url:, topology_url:, collector_definitions_file: ENV["COLLECTOR_DEFINITIONS_FILE"])
+      def initialize(sources_api:, topology_api:, collector_definitions_file: ENV["COLLECTOR_DEFINITIONS_FILE"])
         @collector_definitions_file = collector_definitions_file || TopologicalInventory::Orchestrator.root.join("config/collector_definitions.yaml")
 
-        @logger       = ManageIQ::Loggers::Container.new
-        @sources_url  = sources_url
-        @topology_url = topology_url
+        @logger = ManageIQ::Loggers::Container.new
+
+        @sources_api = sources_api
+        @sources_internal_api = URI.parse(sources_api).tap { |uri| uri.path = "/internal/v1.0" }.to_s
+
+        @topology_api = topology_api
+        @topology_internal_api = URI.parse(topology_api).tap { |uri| uri.path = "/internal/v0.0" }.to_s
       end
 
       def run
@@ -100,19 +102,19 @@ module TopologicalInventory
       end
 
       def sources_api_url_for(path)
-        File.join(sources_url, ENV["PATH_PREFIX"].to_s, ENV["APP_NAME"].to_s, SOURCES_API_VERSION, path)
+        File.join(sources_api, path)
       end
 
       def sources_internal_url_for(path)
-        File.join(sources_url, "internal", "v1.0", path)
+        File.join(sources_internal_api, path)
       end
 
       def topology_api_url_for(path)
-        File.join(topology_url, ENV["PATH_PREFIX"].to_s, ENV["APP_NAME"].to_s, TOPOLOGY_API_VERSION, path)
+        File.join(topology_api, path)
       end
 
       def topology_internal_url_for(path)
-        File.join(topology_url, "internal", "v0.0", path)
+        File.join(topology_internal_api, path)
       end
 
       def each_resource(url, tenant_account = ORCHESTRATOR_TENANT, &block)
--- a/lib/topological_inventory/orchestrator/worker.rb
+++ b/lib/topological_inventory/orchestrator/worker.rb
@@ -38,8 +38,8 @@ module TopologicalInventory
           require "uri"
 
           URI::HTTP.build(
-            :host => ENV["TOPOLOGICAL_INVENTORY_API_SERVICE_HOST"],
-            :port => ENV["TOPOLOGICAL_INVENTORY_API_SERVICE_PORT"],
+            :host => ENV["SOURCES_HOST"],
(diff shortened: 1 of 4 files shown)


lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -38,8 +38,8 @@ def api_base_url
           require "uri"
 
           URI::HTTP.build(
-            :host => ENV["TOPOLOGICAL_INVENTORY_API_SERVICE_HOST"],
-            :port => ENV["TOPOLOGICAL_INVENTORY_API_SERVICE_PORT"],
+            :host => ENV["SOURCES_API_SERVICE_HOST"],
+            :port => ENV["SOURCES_API_SERVICE_PORT"],
```
carbonin: This will need a config change because the sources service isn't in the same namespace.

Also based on [the existing env vars](https://github.com/RedHatInsights/e2e-deploy/blob/master/env/ci.yml#L6-L15) I think the convention would be `SOURCES_HOST` and `SOURCES_PORT`

agrare: :+1: done

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -104,15 +99,20 @@ def collectors_from_sources_api
         hash
       end
 
-      def internal_url_for(path, query = nil)
-        URI.parse(api_base_url).tap do |uri|
-          uri.path = File.join("/internal/v0.0/", path)
-          uri.query = query if query
-        end.to_s
+      def sources_api_url_for(path)
+        File.join(sources_url, ENV["PATH_PREFIX"].to_s, ENV["APP_NAME"].to_s, SOURCES_API_VERSION, path)
```
carbonin: The `APP_NAME` value shouldn't be the same for sources and topo. I think we can probably hardcode these here like we do for the version. They don't seem likely to change at this point.

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -67,16 +57,21 @@ def make_openshift_match_database
...
-          each_resource(url_for("source_types"), tenant) do |source_type|
-            collector_definition = collector_definitions[source_type["name"]]
-            next if collector_definition.nil?
-            each_resource(url_for("source_types/#{source_type["id"]}/sources"), tenant) do |source|
-              next unless endpoint = get_and_parse(url_for("sources/#{source["id"]}/endpoints"), tenant)["data"].first
-              next unless authentication = get_and_parse(url_for("authentications?resource_type=Endpoint&resource_id=#{endpoint["id"]}"), tenant)["data"].first
-              auth = authentication_with_password(authentication["id"], tenant)
-              yield source, endpoint, auth, collector_definition
-            end
+          each_resource(topology_api_url_for("sources"), tenant) do |source_stub|
```
carbonin: I think `topo_source` (or something) would be more clear to me here rather than `source_stub`.

agrare: :+1: Done

lib/topological_inventory/orchestrator/worker.rb
agrare: Hm you're right, let me just move building the api path out of this and provide a full url to the sources and topology api, since hard-coding these to "sources" and "topological-inventory" would break the ability to run locally.

spec/worker_spec.rb
```diff
@@ -47,23 +44,30 @@
...
       EOJ
     end
 
-    let(:source_types_2_sources_response) do
+    let(:sources_1_response) do
       <<~EOJ
         {
-          "links": {},
-          "data": []
+          "id":"1","source_type_id":"1","name":"mock-source","uid":"cacebc33-1ed8-49d4-b4f9-713f2552ee65","tenant":"#{user_tenant_account}"
```
carbonin: Does the sources API not have paging?

spec/worker_spec.rb
```diff
@@ -104,20 +108,22 @@
...
-      stub_rest_get("http://example.com:8080/v0.1/sources/2/endpoints", user_tenant_header, sources_2_endpoints_response)
-      stub_rest_get("http://example.com:8080/v0.1/authentications?resource_type=Endpoint&resource_id=1", user_tenant_header, endpoints_1_authentications_response)
-      stub_rest_get("http://example.com:8080/internal/v0.0/authentications/1?expose_encrypted_attribute[]=password", user_tenant_header, {"username" => "USER", "password" => "PASS"}.to_json)
-
-      expect(RestClient).not_to receive(:get).with("http://example.com:8080/v0.1/authentications?resource_type=Endpoint&resource_id=8", any_args)
-      expect(RestClient).not_to receive(:get).with("http://example.com:8080/v0.1/authentications?resource_type=Endpoint&resource_id=9", any_args)
-      expect(RestClient).not_to receive(:get).with("http://example.com:8080/internal/v0.0/authentications/8?expose_encrypted_attribute[]=password", any_args)
-      expect(RestClient).not_to receive(:get).with("http://example.com:8080/internal/v0.0/authentications/9?expose_encrypted_attribute[]=password", any_args)
+      instance = subject
```
carbonin: Can we just replace the instance usages with subject to avoid this weird line?

spec/worker_spec.rb
agrare: It does, but this is a get on an id not the collection

spec/worker_spec.rb
agrare: :+1: yeah

spec/worker_spec.rb
```diff
@@ -104,22 +108,22 @@
...
+      stub_rest_get("http://example.com:8080/api/sources/v1.0/sources/2", user_tenant_header, sources_2_response)
+      stub_rest_get("http://example.com:8080/api/sources/v1.0/sources/2/endpoints", user_tenant_header, sources_2_endpoints_response)
+      stub_rest_get("http://example.com:8080/api/sources/v1.0/endpoints/1/authentications", user_tenant_header, endpoints_1_authentications_response)
+      stub_rest_get("http://example.com:8080/internal/v1.0/authentications/1?expose_encrypted_attribute[]=password", user_tenant_header, {"username" => "USER", "password" => "PASS"}.to_json)
+
+      expect(RestClient).not_to receive(:get).with("http://example.com:8080/api/sources/v1.0/endpoints/8/authentications", any_args)
+      expect(RestClient).not_to receive(:get).with("http://example.com:8080/api/sources/v1.0/endpoints/9/authentications", any_args)
+      expect(RestClient).not_to receive(:get).with("http://example.com:8080/internal/v1.0/authentications/8?expose_encrypted_attribute[]=password", any_args)
+      expect(RestClient).not_to receive(:get).with("http://example.com:8080/internal/v1.0/authentications/9?expose_encrypted_attribute[]=password", any_args)
```
agrare: @carbonin what are these testing?  I didn't see any endpoints with id 8/9 being created

PR #17, opened 2019-04-26
Lower the cpu request and limit for collector deployment configs

carbonin:
We can now request under 100m for CPU as the Limit Ranges in our
namespaces were updated.

@agrare @Ladas I chose these numbers based on what I can see in the CI env. Do you think we'll get much higher than this?

cc @gtanzillo


carbonin:
I created the limit by doubling the largest value I found for a collector CPU in the CI environment, but I imagine we don't expect them to have particularly large spikes, right?

PR #22, opened 2019-06-16
Add a smoke tests groovy - orchestrator

shalomnaim1:
This PR has to be reviewed by @bsquizz

### DIFF
--- a/Jenkinsfile-smoke.groovy
+++ b/Jenkinsfile-smoke.groovy
@@ -14,7 +14,7 @@ if (env.CHANGE_ID) {
         // the service-set/component for this app in e2e-deploy "templates"
         ocDeployerComponentPath: "topological-inventory/topological-inventory-orchestrator",
         // the service sets to deploy into the test environment
-        ocDeployerServiceSets: "platform-mq, sources, topological-inventory",
+        ocDeployerServiceSets: "platform-mq,sources,topological-inventory",
         // the iqe plugins to install for the test
         iqePlugins: ["iqe-topology_inventory-plugin"],
         // the pytest marker to use when calling `iqe tests all`
--- a/Jenkinsfile-smoke.groovy
+++ b/Jenkinsfile-smoke.groovy
@@ -1,0 +1,24 @@
+/*
+ * Requires: https://github.com/RedHatInsights/insights-pipeline-lib
+ */
+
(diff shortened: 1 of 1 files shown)


Jenkinsfile-smoke.groovy
```diff
@@ -0,0 +1,24 @@
...
+// this 'if' statement makes sure this is a PR, so we don't run smoke tests again
+// after code has been merged into the stable branch.
+if (env.CHANGE_ID) {
+    runSmokeTest (
+        // the service-set/component for this app in e2e-deploy "buildfactory"
+        ocDeployerBuilderPath: "topological-inventory/topological-inventory-orchestrator",
+        // the service-set/component for this app in e2e-deploy "templates"
+        ocDeployerComponentPath: "topological-inventory/topological-inventory-orchestrator",
+        // the service sets to deploy into the test environment
+        ocDeployerServiceSets: "platform-mq, sources, topological-inventory",
```
bsquizz: You'll need to remove the white space: `platform-mq,sources,topological-inventory`

shalomnaim1: fixed

carbonin:
I don't think we need this in every repo. @bsquizz I think one should be sufficient, right?

If we do this everywhere then we'll just end up running the same set of tests with the same set of images something like 8 times, right? I would think we would be fine if we just gate all of the PRs to `stable` on the smoke tests running on the `-api` repo.

bsquizz:
> I don't think we need this in every repo. @bsquizz I think one should be sufficient, right?
> 
> If we do this everywhere then we'll just end up running the same set of tests with the same set of images something like 8 times, right? I would think we would be fine if we just gate all of the PRs to `stable` on the smoke tests running on the `-api` repo.

I chatted with @carbonin about this a bit more to fill him in on what we're shooting for... I think dev & QE need to talk a bit more about what and how they want to test in a smoke run, keeping in mind that the smoke tests have direct access to the pods deployed in the smoke environment, so they can do things like talk directly to kafka or a postgres DB to "seed" data or "read/validate" data. It may be the case that in some repos we can get away with only deploying/testing certain pieces of the topo-inventory set instead of deploying the whole thing repeatedly. And we may be able to seed the DB instead of using a real external provider all of the time.

PR #24, opened 2019-06-21
Add a Metrics scaler

bdunne:
Need to do some testing in a deployed environment


lib/topological_inventory/orchestrator/metric_scaler.rb
```diff
@@ -0,0 +1,88 @@
...
+            total_consumed += h[current_metric_name].to_f
+            total_max      += h[max_metric_name].to_f
+          end
+
+          current_usage_pct = (total_consumed.to_f / total_max.to_f) * 100
+          deviation_pct = current_usage_pct - target_usage_pct
+
+          logger.info("#{dc.metadata.name} consuming #{total_consumed} of #{total_max}, #{current_usage_pct}%")
+
+          next if deviation_pct < scale_threshold_pct # Within tolerance
```
carbonin: I think this should be an absolute value of the deviation_pct as it can be negative. Otherwise the tolerance wouldn't apply when we scale down.

carbonin:
@bdunne were are you at with this? Have you tested it live yet? The code looks good to me.

bdunne:
> Have you tested it live yet?

Not yet.  I'll mark it as WIP until then.

bdunne:
@carbonin Please re-review

lib/topological_inventory/orchestrator/metric_scaler.rb
```diff
@@ -0,0 +1,186 @@
...
+          deviation.positive? ? count += 1 : count -= 1
+          count.clamp(@min_replicas, @max_replicas)
+        end
+
+        def pod_ips
+          endpoint = object_manager.get_endpoint(deployment_config_name)
+          endpoint.subsets.flat_map { |s| s.addresses.collect { |a| a[:ip] } }
+        end
+
+        class FixedLengthArray
```
carbonin: So, normally I'm not particular pedantic about performance ... but this is going to be used constantly.

What you're looking for here is a [circular buffer](https://en.wikipedia.org/wiki/Circular_buffer). What you have implemented currently will create a new internal array object each time we insert a new value. It would be much better to use the same object internally.

[Here's](https://github.com/alindeman/circular_queue/blob/master/lib/circular_queue.rb) a gem that implements a queue with the concept, but the issue is that it assumes readers an writers (as is usually the case with a buffer queue). You would want to just drop the last item when enqueueing.

carbonin: I'm fine if you find something that accomplishes the task or implement your own, whatever seems best.

lib/topological_inventory/orchestrator/metric_scaler.rb
```diff
@@ -0,0 +1,186 @@
...
+        (@cache.keys - dc_names).each { |name| @cache.delete(name).stop }
+        # currently configured
+        dc_names.each                 { |name| @cache[name].scale_to_desired_replicas }
+
+        logger.info("#{self.class.name}##{__method__} Complete")
+      end
+
+      private
+
+      class Watcher
```
carbonin: I think it would be worthwhile to move this out into a separate file. This one is a bit unwieldy.

bin/metrics_scaler
```diff
@@ -0,0 +1,23 @@
...
+end
+
+require "topological_inventory/orchestrator/application_metrics"
+require "topological_inventory/orchestrator/metric_scaler"
+
+metrics = TopologicalInventory::Orchestrator::ApplicationMetrics.new(args[:metrics_port])
+
+Signal.trap("TERM") do
+  metrics.stop_server
+  exit
```
carbonin: Is there any additional cleanup we want to do before we exit. Are we okay leaving the deployments scaled to whatever value when we exit or do we want to set them to some default?

lib/topological_inventory/orchestrator/metric_scaler.rb
```diff
@@ -0,0 +1,186 @@
...
+          object_manager.scale(deployment_config_name, desired_count)
+
+          # Wait for scaling to complete in Openshift
+          sleep(1) until pod_ips.length == desired_count
+          logger.info("Scaling #{deployment_config_name} complete")
+        end
+
+        private
+
+        def moving_usage
```
carbonin: This name is ... not great. Maybe `metric_list`? The name just doesn't convey that it's an array so the line `moving_usage.average` sounds weird.

bin/metrics_scaler
bdunne: I don't think we should change any scaling on exit.

PR #25, opened 2019-06-25
Fix pagination issue

bdunne:
Fixes: PnT JIRA (TPINVTRY-461)

### DIFF
--- a/lib/topological_inventory/orchestrator/worker.rb
+++ b/lib/topological_inventory/orchestrator/worker.rb
@@ -132,7 +132,12 @@ module TopologicalInventory
         resources = paging ? response["data"] : response
         resources.each { |i| yield i }
 
-        each_resource(response.fetch_path("links", "next"), tenant_account, &block) if paging
+        return unless paging
+        next_page_link = response.fetch_path("links", "next")
+        return unless next_page_link
+        next_url = URI.parse(url).merge(next_page_link).to_s
+
+        each_resource(next_url, tenant_account, &block)
       end
 
       def get_and_parse(url, tenant_account = ORCHESTRATOR_TENANT)
--- a/spec/worker_spec.rb
+++ b/spec/worker_spec.rb
@@ -154,12 +154,15 @@ describe TopologicalInventory::Orchestrator::Worker do
 
   describe "#each_resource" do
     it "works with paginated responses" do
-      url_1 = "http://example.com:8080/1"
-      url_2 = "http://example.com:8080/2"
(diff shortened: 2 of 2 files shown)


lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -132,7 +132,9 @@ def each_resource(url, tenant_account = ORCHESTRATOR_TENANT, &block)
         resources = paging ? response["data"] : response
         resources.each { |i| yield i }
 
-        each_resource(response.fetch_path("links", "next"), tenant_account, &block) if paging
+        return unless paging && next_page_link = response.fetch_path("links", "next")
```
carbonin: I think `next_page_path` would make the issue more explicit in the code here. The fact that we're only getting a path rather than an actual link is the problem we're trying to solve.

Also, multi condition `unless` one-liners are hard to parse.

Maybe this would be easier to read if we moved the assignment to the line above?

```ruby
next_page_path = response.fetch_path("links", "next")
return unless paging && next_page_path

next_url = URI.parse(url).merge(next_page_path).to_s
each_resource(next_url, tenant_account, &block)
```

and even then I still prefer the `if` version:
```ruby
return if !paging || next_page_path.nil?
```

bdunne: > I think next_page_path would make the issue more explicit in the code here.

I didn't want to use `path` in the variable name, because it's not just a path, it's a relative URL.  In our case it's `path` and `query`.

I think `return unless paging && next_page_link` is clearer than `return if !paging || next_page_path.nil?`.  If you prefer, I can change to something like this maybe?
```ruby
return unless paging
return unless next_page_link
```

bdunne: Also, I was thinking about dropping the whole conditional around paging, but didn't want to do that in this PR.  What do you think?  Both topo and sources should be paginated.

carbonin: > I didn't want to use path in the variable name, because it's not just a path, it's a relative URL. In our case it's path and query.

Makes sense. I'm okay with that logic.

> I think `return unless paging && next_page_link` is clearer than `return if !paging || next_page_path.nil?`

I'm okay with this too as long as the assignment makes its way out of the conditional.

> Also, I was thinking about dropping the whole conditional around paging, but didn't want to do that in this PR. What do you think? Both topo and sources should be paginated.

Both regular APIs are, but I added it for the internal APIs for tenants and authentications which I do not believe are paginated.

PR #30, opened 2019-07-17
Rubocop

slemrmartin:
Rubocop

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,8 @@
 .bundle
 Gemfile.lock
+
+# generated rubocop file
+.rubocop-https*
+
+## Environment normalization:
+bundler.d/*
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,4 @@
 Gemfile.lock
 
 # generated rubocop file
-.rubocop-https*
-
-## Environment normalization:
-bundler.d/*
+.rubocop*
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@
 Gemfile.lock
 
 # generated rubocop file
-.rubocop*
+.rubocop-*
--- a/.rubocop_local.yml
+++ b/.rubocop_local.yml
@@ -1,0 +1,3 @@
+AllCops:
+  Exclude:
+  - db/schema.rb
--- a/.rubocop_local.yml
+++ b/.rubocop_local.yml
@@ -1,3 +1,0 @@
-AllCops:
-  Exclude:
-  - db/schema.rb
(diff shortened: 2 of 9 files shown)


slemrmartin:
@bdunne pls :eyes:

.rubocop_local.yml
```diff
@@ -0,0 +1,3 @@
+AllCops:
+  Exclude:
+  - db/schema.rb
```
bdunne: We don't have a `db` directory in this repo

.gitignore
```diff
@@ -1,2 +1,8 @@
 .bundle
 Gemfile.lock
+
+# generated rubocop file
+.rubocop-https*
+
+## Environment normalization:
+bundler.d/*
```
bdunne: We don't have a `bundler.d` directory in this repo

.gitignore
```diff
@@ -1,2 +1,8 @@
 .bundle
 Gemfile.lock
+
+# generated rubocop file
+.rubocop-https*
```
bdunne: We ignore `.rubocop-*` in the plugin generator

.rubocop_local.yml
slemrmartin: ok, I copied it from other repo

.gitignore
slemrmartin: What it means? I added it, because `.rubocop-https---raw-githubusercontent-com-ManageIQ-guides-master--rubocop-base-yml` this file is always autogenerated locally

slemrmartin: btw core, persister, api & collectors doesn't ignore `.rubocop-*` (which includes `.rubocop-cc.yml` and `.rubocop_local.yml`

slemrmartin: ah, sorry, I got it, these files are empty now, removing

bdunne: > What it means? I added it, because .rubocop-https---raw-githubusercontent-com-ManageIQ-guides-master--rubocop-base-yml this file is always autogenerated locally

See:
https://github.com/ManageIQ/manageiq/blob/master/lib/generators/manageiq/plugin/templates/.gitignore#L1
https://github.com/ManageIQ/manageiq/blob/master/lib/generators/manageiq/plugin/templates/.rubocop.yml
https://github.com/ManageIQ/manageiq/blob/master/lib/generators/manageiq/plugin/templates/.rubocop_cc.yml
https://github.com/ManageIQ/manageiq/blob/master/lib/generators/manageiq/plugin/templates/.rubocop_local.yml

> btw core, persister, api & collectors doesn't ignore .rubocop-* (which includes .rubocop-cc.yml and .rubocop_local.yml

I think @Fryguy forgot to run some repo setup scripts.  BTW, `.rubocop-*` does not match `.rubocop_cc.yml` or `.rubocop_local.yml` and those files should exist even if they are blank.

slemrmartin: ok, it should be ok now

PR #31, opened 2019-07-29
Multisource collectors support

slemrmartin:
## How it works

This PR reads sources from API and assigns each source to ConfigMap, Secret and DeploymentConfig. Each collector can collect more sources (of the same SourceType) now.

Each SourceType (allowed) has assigned `sources_per_collector` value (yaml file `config/default.yml`).

Each Source is associated with one Config map. ConfigMap can group sources of the same type to the amount up to `sources_per_collector`. 

Each ConfigMap is identified by UUID and labeled by source type.
For each ConfigMap there is created Secret and DeploymentConfig. 

ConfigMap and Secret are mapped to filesystem to `<app_dir>/config/custom.yml`, `<app_dir>/secrets/credentials`. These values are periodically read by collector (so adding/removing of source doesn't need restart of pod).

Single-source(old) deployment configs will be deleted automatically after restarting orchestrator with this code (detected by label `topological-inventory/collector_digest`)

## Self-recovery
If any of config-map, secret or deployment config is deleted manually (outside of orchestrator), orchestrator detects it and creates missing openshift objects

## Filtering in openshift (labels)

- label `tp-inventory/config_uid` (Unique label)
  -  each ConfigMap and assocaited Secret and DeploymentConfig
- label `tp-inventory/source-type` 
  - filtering by source type
- label `tp-inventory/collectors-config-map`
  - all config maps 
- label `tp-inventory/collector`
  - all deployment configs
- label `tp-inventory/collectors-secret`
  - all secrets
- ???
  - Source can be identified only by config maps's data

## Links

**depends on**
- [x] https://github.com/ManageIQ/topological_inventory-ansible_tower/pull/22
- [x] https://github.com/ManageIQ/topological_inventory-amazon/pull/36
- [x] https://github.com/ManageIQ/topological_inventory-ansible_tower/pull/23
- [x] https://github.com/ManageIQ/topological_inventory-openshift/pull/98

**related**
- [ ] https://github.com/RedHatInsights/e2e-deploy/pull/580

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1,0 +1,5 @@
+collectors:
+  sources_per_collector:
+    amazon: 2
+    ansible-tower: 2
+    openshift: 2
--- a/lib/topological_inventory/orchestrator/logger.rb
+++ b/lib/topological_inventory/orchestrator/logger.rb
@@ -1,0 +1,19 @@
+require "manageiq/loggers"
+
+module TopologicalInventory
+  module Orchestrator
+    class << self
+      attr_writer :logger
+    end
+
+    def self.logger
+      @logger ||= ManageIQ::Loggers::CloudWatch.new
+    end
+
+    module Logging
+      def logger
+        TopologicalInventory::Orchestrator.logger
+      end
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ plugin "bundler-inject", "~> 1.1"
 require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundler-inject") rescue nil
 
 gem "cloudwatchlogger", "~>0.2"
+gem "config", "~> 1.7.2"
 gem "kubeclient", "~>4.0"
 gem "manageiq-loggers", "~>0.4"
 gem "more_core_extensions", "~>3.7.0"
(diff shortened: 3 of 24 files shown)


slemrmartin:
@miq-bot remove-label unmergeable

slemrmartin:
@carbonin @bdunne please review. 

There are some problems to discussion: 
- How many sources for one pod set by default?
- Which collector is processing which source cannot be identified directly by pod's name, because it's array of sources (and openshift object's name size is limited to 64 characters). It can be determined by associated config map only (I considered adding labels for each source (label cannot be array), but it'll create too many labels

lib/topological_inventory/orchestrator/logger.rb
```diff
@@ -0,0 +1,20 @@
...
+
+module TopologicalInventory
+  module Orchestrator
+    class << self
+      attr_writer :logger
+    end
+
+    def self.logger
+      @logger ||= ManageIQ::Loggers::CloudWatch.new
+      @logger
```
bdunne: Is this line necessary? The one above should return the object.

bdunne:
So, there is a lot to look at here.  I really like the restructuring of API objects and Openshift objects, but it also makes it difficult to differentiate the changes that are necessary for multi-collector and the refactoring.  There have been a bunch of things merged recently that require a rebase of this PR, would you be able to split this into a few PRs (maybe starting with the API Objects)?  I think that would be the easiest and quickest way to get this in.

config/default.yml
```diff
@@ -0,0 +1,5 @@
+collectors:
+  sources_per_collector:
```
slemrmartin: @agrare @bdunne how many sources per collector by default?

slemrmartin: I don't know how about CPU, but I can do some benchmark how much memory consumes one collecting of one source, and @Ladas have some benchmark I think, so it'll be something like 
`(max(single-source collector's RAM) - Rails app RAM) * number of sources) + Rails app RAM`

Ladas: I think the threading is still using GIL, correct? So I am pretty sure that can't go over 1 core.

As for memory, yeah the formula looks about right. (We'll need to make sure that all API calls are nicely paginated, so we don't keep much in memory and there are no spikes)

PR #33, opened 2019-08-06
Add basic quota checks before trying to create a DeploymentConfig

bdunne:
Start to TPINVTRY-532

### DIFF
--- a/lib/topological_inventory/orchestrator/object_manager.rb
+++ b/lib/topological_inventory/orchestrator/object_manager.rb
@@ -1,10 +1,17 @@
 require "kubeclient"
+require "more_core_extensions/core_ext/string/iec60027_2"
 
 module TopologicalInventory
   module Orchestrator
     class ObjectManager
       TOKEN_FILE   = "/run/secrets/kubernetes.io/serviceaccount/token".freeze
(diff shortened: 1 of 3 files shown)


lib/topological_inventory/orchestrator/object_manager.rb
```diff
@@ -98,6 +106,39 @@ def manager_uri(path)
...
+        memory_request = quota_status.used["requests.memory"].iec_60027_2_to_i
+        definition.dig(:spec, :template, :spec, :containers).each do |container|
+          memory_request += container.dig(:resources, :requests, :memory).iec_60027_2_to_i
+        end
+        raise(QuotaMemoryRequestExceeded) if memory_request >= quota_status.hard["requests.memory"].iec_60027_2_to_i
+      end
+
+      def cpu_string_to_millicores(input)
+        match = input.match(/(?<number>\d+)(?<suffix>m?)/)
+        match[:suffix].empty? ? (match[:number].to_i * 1000) : match[:number].to_i
```
carbonin: Isn't this logic reversed? If we don't match a suffix then we have cores, right?

bdunne: Nope

PR #34, opened 2019-08-12
Update deployment status

bdunne:
TPINVTRY-532

### DIFF
--- a/lib/topological_inventory/orchestrator/worker.rb
+++ b/lib/topological_inventory/orchestrator/worker.rb
@@ -79,14 +79,14 @@ module TopologicalInventory
             auth = authentication_with_password(authentication["id"], tenant)
             next if auth.nil?
 
-            yield source, endpoint, auth, collector_definition
+            yield source, endpoint, auth, collector_definition, tenant
           end
         end
       end
 
       def collectors_from_sources_api
         hash = {}
-        each_source do |source, endpoint, authentication, collector_definition|
+        each_source do |source, endpoint, authentication, collector_definition, tenant|
           value = {
             "endpoint_host"   => endpoint["host"],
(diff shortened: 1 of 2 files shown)


lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -100,15 +100,16 @@ def collectors_from_sources_api
...
           key = digest(value)
           hash[key] = value
         end
         hash
       end
 
-      def sources_api_url_for(path)
-        File.join(sources_api, path)
+      def sources_api_url_for(*path)
+        File.join(sources_api, *path)
```
agrare: This was a nice change :+1:

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -100,15 +100,16 @@ def collectors_from_sources_api
               "password" => authentication["password"],
               "username" => authentication["username"],
             },
+            "tenant"          => tenant,
```
slemrmartin: @bdunne Why do you add tenant into digest? If it's possible to change tenant for source, then it is still the same provider with the same endpoint and credentials.
Changing of digest means that this is new source from orchestrator's/openshift's point of view => which means kill and redeploy. 
And this is not necessary.

bdunne: I need it to update the source record [here](https://github.com/ManageIQ/topological_inventory-orchestrator/pull/34/files#diff-4743ee12ee7e621468f2e6590de994efR218)

slemrmartin: ah, ok, so you don't need it for digest but for rest api call. 
I won't keep it in #31, tenant will be saved in `Source` instance

PR #37, opened 2019-08-21
Only deploy collectors for supported applications

agrare:
The only applications that depend on topological-inventory data and thus
require collectors to be deployed are catalog and topological-inventory.

If a source isn't enabled for any of these application_types then we
shouldn't deploy collectors for it.

Fixes https://projects.engineering.redhat.com/browse/TPINVTRY-546

### DIFF
--- a/lib/topological_inventory/orchestrator/worker.rb
+++ b/lib/topological_inventory/orchestrator/worker.rb
@@ -73,7 +73,7 @@ module TopologicalInventory
 
             application_types = get_and_parse(sources_api_url_for("sources", source["id"], "application_types"), tenant)
 
-            enabled_applications = application_types&.dig("data")&.map { |app_type| app_type["name"] } || []
+            enabled_applications = application_types["data"].map { |app_type| app_type["name"] }
             next if (SUPPORTED_APPLICATIONS & enabled_applications).empty?
 
             endpoints = get_and_parse(sources_api_url_for("sources", source["id"], "endpoints"), tenant)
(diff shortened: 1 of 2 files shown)


agrare:
We could probably make this more efficient by doing one request to `/applications` per-tenant for the supported application_type_ids but wanted to get started on this.

We can always make performance enhancements as a follow-up

lib/topological_inventory/orchestrator/worker.rb
```diff
@@ -70,6 +71,11 @@ def each_source
 
             next unless (collector_definition = collector_definitions[source_type["name"]])
 
+            application_types = get_and_parse(sources_api_url_for("sources", source["id"], "application_types"), tenant)
+
+            enabled_applications = application_types&.dig("data")&.map { |app_type| app_type["name"] } || []
```
bdunne: Will `application_types` ever be `nil`?

agrare: Ideally that's correct, `get_and_parse` does return `nil` on a `RestClient::NotFound` exception though not sure if that's something we need to handle here

agrare: I could just `next if application_types.nil?` and avoid the `&. &.`, if it is non-nil then it should have the proper `"data"` key, wyt?

bdunne: I think the only time we would get a `RestClient::NotFound` is if the source doesn't exist and [we already account for that](https://github.com/ManageIQ/topological_inventory-orchestrator/pull/37/files#diff-4743ee12ee7e621468f2e6590de994efR68)

agrare: :+1: done

PR #45, opened 2019-08-28
Cascade delete for deployment configs

slemrmartin:
This PR fixes situation when DC of collector contains more versions of deployments (build was updated) and orchestrator tries to delete deployment config.

For deleting of all deployment versions Foreground cascade delete has to be set:
https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/


slemrmartin:
cc @carbonin 
@bdunne please :eyes:

PR #46, opened 2019-08-29
Changed DC deleting policy to Background

slemrmartin:
Foreground policy doesn't work with current master, because DC is marked for deletion and before all deployments are deleted, then orchestrator runs new round and creates new DC with the same name (which is obviously the same object in openshift) - and this leads to some zombie state.

### DIFF
--- a/lib/topological_inventory/orchestrator/object_manager.rb
+++ b/lib/topological_inventory/orchestrator/object_manager.rb
@@ -61,10 +61,10 @@ module TopologicalInventory
 
         scale(name, 0)
         delete_options = Kubeclient::Resource.new(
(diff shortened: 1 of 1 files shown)


slemrmartin:
@Ladas pls merge

slemrmartin:
cc @bdunne

agrare:
Nice @slemrmartin !  This sounds like it fixes https://projects.engineering.redhat.com/browse/TPINVTRY-621 as that is the behavior we were seeing where the deployments were cancelled or complete.

slemrmartin:
@agrare maybe yes, but it primarily fixed https://github.com/ManageIQ/topological_inventory-orchestrator/pull/45 from yesterday

PR #47, opened 2019-08-30
More source's info in Config map

slemrmartin:
Now config map contains only source's UID. 
It'll easier in case of problems to know more about it in ConfigMap's data.

This PR shouldn't change any functionality

### DIFF
--- a/spec/config_map_spec.rb
+++ b/spec/config_map_spec.rb
@@ -5,8 +5,8 @@ describe TopologicalInventory::Orchestrator::ConfigMap do
 
   let(:object_manager) { double('object_manager') }
   let(:openshift_object) { double('openshift_object') }
-  let(:source) { double('source') }
-  let(:source2) { double('source2') }
+  let(:source) { TopologicalInventory::Orchestrator::Source.new({'id' => '1', 'uid' => 'source-1', 'name' => 'Source 1'}, nil, nil, nil, from_sources_api: true) }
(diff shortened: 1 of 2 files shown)


slemrmartin:
@miq-bot assign @bdunne

slemrmartin:
hey @agrare can you look? It extends configmap's data, I think for `oc describe` diagnostics it'll speed up finding source.

spec/config_map_spec.rb
```diff
@@ -15,10 +15,21 @@
 
   before do
     allow(source).to receive(:source_type).and_return(source_type)
-    allow(source).to receive(:[]) { |arg| arg == 'uid' ? 'source-1' : nil }
+    allow(source).to receive(:[]) do |arg|
+      case arg
+      when 'id' then '1'
+      when 'uid' then 'source-1'
+      when 'name' then 'Source 1'
+      end
```
agrare: We should either do either
`let(:source) { SourcesApiClient::Source.new(:source_type => ..., :uid => ..., etc...)`
or
`let(:source) { OpenStruct.new(...) }`

I would argue for the first one, we already depend on SourcesApiClient as a dependency so stubbing everything doesn't reduce inter-dependency it would just cover up a change to SourcesApiClient breaking this repo.  If a dependency change breaks this repo I'd rather find out from a scheduled cron travis run then in production

slemrmartin: @agrare this source is not an `SourcesApiClient::Source`, but orchestrator's Source. But I get the point, I could simulate the Source's structure better there.

slemrmartin: @agrare changed to `Orchestrator::Source`. Btw we don't relay on SourcesApiClient in this repo, it's using RestClient instead.

PR #48, opened 2019-08-30
Removing invalid source log

slemrmartin:
It seems `Source 022abaf2-ac6b-43d1-a057-259f8a4281e6 invalid, no data for digest` is kibana killer, so better to remove this log


slemrmartin:
@miq-bot assign @agrare

agrare:
@slemrmartin we used to skip sources without endpoints or authentications without an issue, I think we should just remove the log instead of commenting it out.

PR #50, opened 2019-09-03
Event driven orchestrator

slemrmartin:
By Kafka's queue `platform.sources.event-stream`

Sources API generates events *create*, *update*, *destroy* on `Source`, `Endpoint`, `Authentication` and `Application`. 

Full sync is started with any of these events.  
Because of UI is updating whole source (Source,Endpoint,Authentication) at the same time, multiple events are generated. Thus orchestrator ignores events which came 5 seconds since first one.

If no event is cought during 1 hour, orchestrator generates `Orchestrator.sync` event which invokes the same as standard event ^^^.

`NO_KAFKA` env variable disables event-driven a switches back to each-10-seconds sync

---

- [x] **depends on** https://github.com/RedHatInsights/e2e-deploy/pull/615

### DIFF
--- a/lib/topological_inventory/orchestrator/event_manager.rb
+++ b/lib/topological_inventory/orchestrator/event_manager.rb
@@ -1,9 +1,14 @@
 module TopologicalInventory
   module Orchestrator
+    # Event Manager subscribes to kafka topic "platform.sources.event-stream"
+    # and listens to messages from Sources API: create/update/destroy
+    # on models used for computing of digest (see Source#digest_values)
+    #
+    # When event is received then sync API -> OpenShift is invoked.
+    # If no event is caught for >= 1 hour, sync is invoked too.
     class EventManager
       include Logging
 
-      ORCHESTRATOR_EVENT_NAME = "Orchestrator.sync".freeze
       SYNC_EVENT_INTERVAL = 1.hour
       SKIP_SUBSEQUENT_EVENTS_DURATION = 5.seconds
 
@@ -22,7 +27,7 @@ module TopologicalInventory
 
       def run!
         Thread.new { listener }
-        loop { publisher }
+        loop { scheduled_sync }
       end
 
       private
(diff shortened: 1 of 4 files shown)


slemrmartin:
@agrare please :eyes: 
@miq-bot assign @agrare

slemrmartin:
@miq-bot add-label enhancement

lib/topological_inventory/orchestrator/event_manager.rb
```diff
@@ -0,0 +1,129 @@
...
+        @events = %w[Source Endpoint Authentication Application].collect do |model|
+          %W[#{model}.create #{model}.update #{model}.destroy]
+        end.flatten
+        @events << ORCHESTRATOR_EVENT_NAME
+        @events
+      end
+
+      def publish_sync_event
+        publish_opts = {
+          :service => queue_name,
```
agrare: This queue is actually monitored by a lot of other apps, and publishing something there for use internally by the orchestrator really isn't good.

lib/topological_inventory/orchestrator/event_manager.rb
```diff
@@ -0,0 +1,129 @@
...
+        self.worker = worker
+        self.event_semaphore = Mutex.new
+        self.sync_semaphore  = Mutex.new
+        self.last_event_at = nil
+        self.publisher_sleep_time = 0
+      end
+
+      def run!
+        Thread.new { listener }
+        loop { publisher }
```
agrare: Instead of this publishing a message to the topic, I think you could use `messaging_client.stop` which should break the loop every `SYNC_EVENT_INTERVAL`

lib/topological_inventory/orchestrator/event_manager.rb
slemrmartin: removed

lib/topological_inventory/orchestrator/event_manager.rb
slemrmartin: I found more straightforward solution, sync is invoked directly and messaging client isn't needed to stop.

lib/topological_inventory/orchestrator/event_manager.rb
```diff
@@ -0,0 +1,128 @@
...
+
+      def self.run!(worker)
+        manager = new(worker)
+        manager.run!
+      end
+
+      def initialize(worker)
+        self.worker = worker
+        self.event_semaphore = Mutex.new
+        self.sync_semaphore  = Mutex.new
```
agrare: I'd really like to avoid coordinating a multi-mutex dance between threads if at all possible

If we drop the "only do a full sync if there wasn't an event in the last hour" then we don't have to synchronize on last_event_at.

And if you had only a single thread doing the sync openshift with sources then you don't need the second one.  This could be done a few ways but one way could be to have the sync thread listen to an in-memory queue that the kafka thread or the scheduled_sync thread push to.  This is how most of our MIQ event catchers work.

slemrmartin: well, I'm, not sure if I understand. 
1) `last_event_at` is there primarily for skipping multievents generated by Sources API. 
2) if we drop the no-event sync, then we are actually saying: "I believe nothing wrong in another application can happen. Never"...emm, why?
3) both mutexes have different usage, `event_semaphore` sets last sync and handles multievents, `sync_semaphore` handles events from different sources. 
4) code isn't wrong only because it uses threads :)

agrare: > 1. last_event_at is there primarily for skipping multievents generated by Sources API.

That's fine, I'm saying accessing last_event_at from both threads is what is making that mutex necessary

> 2. if we drop the no-event sync, then we are actually saying: "I believe nothing wrong in another application can happen. Never"...emm, why?

Huh?  That's not what I'm saying at all.

> 3. both mutexes have different usage, event_semaphore sets last sync and handles multievents, sync_semaphore handles events from different sources.

Yeah I got that, I'm saying that they might not be necessary if we structure the code differently

> 4. code isn't wrong only because it uses threads :)

That's also not what I'm saying.  Sharing variables between threads is however more complicated and more likely that in 2 years someone who doesn't understand threading as well will screw it up.

In fact I think _adding_ a thread but removing any shared variables between threads would be better

lib/topological_inventory/orchestrator/event_manager.rb
```diff
@@ -0,0 +1,126 @@
...
+            process_event(event[:event_name], event[:model_id])
+          end
+
+          sleep(POLL_TIME)
+        end
+      end
+
+      def process_event(event_name, model_id = nil)
+        # Just one sync at the same time
+        sync_semaphore.synchronize do
```
agrare: Now that process_event is only called from one place we don't need this lock anymore?

agrare:
Looks good @slemrmartin !  I think we can drop that one mutex now otherwise looks good

lib/topological_inventory/orchestrator/event_manager.rb
slemrmartin: right, I overcombined it, just a moment...

PR #51, opened 2019-09-04
GitIgnore for bundler.d

slemrmartin:
bundler.d dir and more common ignored dirs/files


slemrmartin:
@miq-bot assign @Ladas

PR #52, opened 2019-09-10
Nicer names of resources

slemrmartin:
ConfigMaps, DeploymentConfigs and Secrets can have source type in name at least.

Before deploying this, orchestrator should be scaled to 0 and old resources deleted:
- `oc get configmaps | awk '{print $1}' | grep 'tp-inventory-' | xargs oc delete configmap`
- `oc get secrets | awk '{print $1}' | grep 'tp-inventory-' | xargs oc delete secret`
- `oc get dc | awk '{print $1}' | grep 'tp-inventory-' | xargs oc delete dc`


slemrmartin:
@agrare please :eyes:

agrare:
Nice @slemrmartin !

PR #61, opened 2019-12-02
Sync times configurable

slemrmartin:
Event sync poll time and scheduled full sync are now configurable by yaml config

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -4,3 +4,6 @@ collectors:
     ansible-tower: 2
     azure: 2
     openshift: 2
+sync:
+  poll_time_seconds: 10
+  scheduled_event_hours: 1
--- a/lib/topological_inventory/orchestrator/event_manager.rb
+++ b/lib/topological_inventory/orchestrator/event_manager.rb
@@ -9,9 +9,6 @@ module TopologicalInventory
     class EventManager
       include Logging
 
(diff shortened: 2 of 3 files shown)


config/default.yml
```diff
@@ -4,3 +4,6 @@ collectors:
     ansible-tower: 2
     azure: 2
     openshift: 2
+sync:
+  poll_time_seconds: 10 # seconds
+  scheduled_event_hours: 1 # hours
```
bdunne: Do you think we will want this to be more frequent in the future?  We usually specify numbers in the base unit (time in seconds)

config/default.yml
```diff
@@ -4,3 +4,6 @@ collectors:
     ansible-tower: 2
     azure: 2
     openshift: 2
+sync:
+  poll_time_seconds: 10 # seconds
```
bdunne: :+1: the suffix is nice to avoid confusion.  I think it makes the comment unnecessary.

config/default.yml
slemrmartin: I think we will want this to be much less frequent, like once per day/week

slemrmartin: @bdunne it would be good to know how many sources are expected in production. If something like milions, then full sync should be disabled or only on demand.

bdunne: I don't think we have any estimates yet.  We''ll have to adjust as load increases.

PR #62, opened 2019-12-06
Targeted Update

slemrmartin:
Event in kafka queue (published by Sources API) targets one Source (or related Endpoint/Authentication/Application).
This PR requests and updates only necessary data for create/update/delete of one Source

### DIFF
--- a/lib/topological_inventory/orchestrator/api.rb
+++ b/lib/topological_inventory/orchestrator/api.rb
@@ -16,7 +16,7 @@ module TopologicalInventory
       end
 
       def each_tenant
-        each_resource(topology_internal_url_for("tenants")) { |tenant| yield tenant }
+        each_resource(topology_internal_url_for("tenants")) { |tenant| yield tenant["external_tenant"] }
       end
 
       def each_source_type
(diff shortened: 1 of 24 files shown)


slemrmartin:
@agrare please :eyes:

lib/topological_inventory/orchestrator/api.rb
```diff
@@ -123,9 +123,11 @@ def tenant_header(tenant_account)
 
       # Set of ids for supported applications
       def supported_application_type_ids
+        return @supported_application_type_ids if @supported_application_type_ids.present?
```
agrare: If we cache this and never reload it, we wouldn't ever pick up changes to the supported applications until the orchestrator was restarted right?

Maybe we should cache_with_timeout this?

slemrmartin: @agrare is there some trick how to rewrite object method to cache_with_timeout's class method?
Because it's using other non-class methods and it seems that cache_with_timeout doesn't receive any arguments.

agrare: We'd have to rewrite this to be a class method (and anything it depends on) which I think is okay because none of this is tenant-level it is all service-level.

agrare: Arg I just noticed that the sources_api url is an instance var, we could either make this a class var or create our own "instance method" version of cache_with_timeout

slemrmartin: resolved by class var

lib/topological_inventory/orchestrator/api.rb
```diff
@@ -121,8 +124,18 @@ def tenant_header(tenant_account)
...
 
+      def supported_application_type_ids(timeout = 300)
+        @@cache_lock.synchronize do
+          if @@cache_start.nil? || @@cache_start <= timeout.seconds.ago
+            @@supported_application_type_ids = load_supported_application_type_ids
+            @@cache_start = Time.now
+          end
+        end
+        @@supported_application_type_ids
+      end
```
agrare: If we were going to do this at the class level it was so that we could use `cache_with_timeout`, if we're going to reimplement the syncronize I don't think it has to be a class level cache but we can reevaluate in a follow-up

slemrmartin: we need to save it at class level because each sync round creates it's own API object, but we need to use API calls, which are implemented at object level, which doesn't have equivalent for `cache_with_timeout`

PR #64, opened 2020-01-07
Updated API calls to be identified as org_admin calls

abellotti:
- identity header used now have the is_org_admin flag set as true.

### DIFF
--- a/lib/topological_inventory/orchestrator/api.rb
+++ b/lib/topological_inventory/orchestrator/api.rb
@@ -118,7 +118,7 @@ module TopologicalInventory
       end
 
       def tenant_header(tenant_account)
-        {"x-rh-identity" => Base64.strict_encode64({"identity" => {"account_number" => tenant_account}}.to_json)}
+        {"x-rh-identity" => Base64.strict_encode64({"identity" => {"account_number" => tenant_account, "user" => { "is_org_admin" => true }}}.to_json)}
       end
 
       # Set of ids for supported applications
--- a/spec/helpers/mock_data.rb
+++ b/spec/helpers/mock_data.rb
@@ -20,11 +20,11 @@ module MockData
   end
 
   def user_tenant_header
-    {"x-rh-identity" => "eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjEyMzQ1In19"}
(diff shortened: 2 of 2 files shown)


lib/topological_inventory/orchestrator/api.rb
```diff
@@ -118,7 +118,7 @@ def get_and_parse(url, tenant_account = Worker::ORCHESTRATOR_TENANT)
       end
 
       def tenant_header(tenant_account)
-        {"x-rh-identity" => Base64.strict_encode64({"identity" => {"account_number" => tenant_account}}.to_json)}
+        {"x-rh-identity" => Base64.strict_encode64({"identity" => {"account_number" => tenant_account, "user" => { "is_org_admin" => true }}}.to_json)}
```
carbonin: We actually query as all the tenants in the topo app. Will this ever be a problem? Is there ever a chance that a tenant we  are querying as will not be an org admin? What would happen?

agrare: I guess every tenant must have at least one org-admin right?

carbonin: Probably? But is every account number we send guaranteed to be an org admin? And if we say an account is an org admin when it isn't, what are the consequences?

Or even more generally, why are we adding this at all?

agrare: > But is every account number we send guaranteed to be an org admin

I don't think the account_number is for a user, it is for the org.  I don't know that it makes sense to ask if an org is an org-admin, only users can be or not be org-admins.

> Or even more generally, why are we adding this at all?

Yeah this is a good question, the reason is that we got a request to completely disable sources for non-org-admins so we need to update all of our internal callers as well.  Maybe the answer is to use the internal API instead of pretending to be an org-admin? IDK

carbonin: We are using the topo internal url to query for the tenant list. So if topo will still contain tenants that are not an org admin then we could run into issues there.

Additionally, when we query without a specific tenant in mind (for example, querying sources for source types) we are using tenant `system_orchestrator`. I would also make sure that we are not going to try to query out to some external service with that (obviously invalid) tenant id.

carbonin: Sorry, I thought org_admin was an attribute of the tenant when it's really related to the user and we just need to fake it here. This is fine.

PR #66, opened 2020-01-23
Removing useless source type logs

slemrmartin:
Only supported SourceTypes should be logged - currently sources_per_collectors settings are logged for ManageIQ's Sources and this is confusing.

### DIFF
--- a/lib/topological_inventory/orchestrator/worker.rb
+++ b/lib/topological_inventory/orchestrator/worker.rb
@@ -89,8 +89,10 @@ module TopologicalInventory
         @source_types_by_id = {}
 
         @api.each_source_type do |attributes|
-          @source_types_by_id[attributes['id']] = SourceType.new(attributes)
-          logger.debug("Loaded Source type: #{attributes['name']} | #{@source_types_by_id[attributes['id']].sources_per_collector} sources per config map")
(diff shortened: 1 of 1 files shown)


agrare:
Could we maybe remove all of this logging?  Not sure if it is still actively helpful and it seems weird to log only some types but not others

slemrmartin:
it could be useful to know if settings are loaded correctly (especially because it's from config map which can be provided by user (qa)).
On the other side logging of how many ovirt/vmware sources we can have doesn't make sense because it's not related to topological inventory.

agrare:
> unfortunatelly supported source types are not defined by Sources API

Aren't we able to get supported_source_types for an application_type?  https://github.com/RedHatInsights/sources-api/blob/master/db/seeds/application_types.yml#L28-L32

agrare:
Anyway I'm good with this I just think we can do better than a hard coded list of supported source_types

slemrmartin:
> > unfortunatelly supported source types are not defined by Sources API
> 
> Aren't we able to get supported_source_types for an application_type? https://github.com/RedHatInsights/sources-api/blob/master/db/seeds/application_types.yml#L28-L32

:open_mouth: didn't know that, I'll add cached API request later

PR #70, opened 2020-03-03
Decrease Towers count in collector

slemrmartin:
**Issue**: https://github.com/RedHatInsights/topological_inventory-ansible_tower/issues/72

Decreasing Sources per collector for Tower to 2. 

Doesn't affect current Sources <-> Collector(ConfigMap) distribution. 
Source will be moved to new collector(config map) if `digest` changes - it means endpoint or authentication values.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1,7 +1,7 @@
 collectors:
   sources_per_collector:
     amazon: 10
-    ansible-tower: 10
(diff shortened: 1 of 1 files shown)


slemrmartin:
@gtanzillo please cherry-pick also to stable (reported by QA)

PR #71, opened 2020-03-04
Add support for the openshift v4 api paths

carbonin:
This is specifically the path that works for deployment configs
which is the only object we interact with in the openshift api

We try the v3 path first, and if that doesn't work we try v4

Fixes https://projects.engineering.redhat.com/browse/TPINVTRY-876

### DIFF
--- a/lib/topological_inventory/orchestrator/object_manager.rb
+++ b/lib/topological_inventory/orchestrator/object_manager.rb
@@ -118,14 +118,14 @@ module TopologicalInventory
       private
 
       def connection
-        @connection ||= raw_connect(manager_uri("/oapi"))
+        @connection ||= detect_openshift_connection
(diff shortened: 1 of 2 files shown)


slemrmartin:
> This is specifically the path that works for deployment configs
> which is the only object we interact with in the openshift api

@carbonin 
Orchestrator works with DeploymentConfigs, ConfigMaps and Secrets
ConfigMaps and Secrets are using different endpoint than DCs in v3 (so I assume also in v4).

Also there are accesses to resources quotas and replication controller objects.

carbonin:
Deployment configs are the only objects that are specific to the openshift API. Note how for the other objects we use the `/api` path (`#kube_connection` method). No other change is needed @slemrmartin

PR #78, opened 2020-03-25
Fixed destroy events when source deleted from db

slemrmartin:
When some Destroy event is processed, their relations in Sources API or Topological API or both needn't exist. 

This PR cover these cases


slemrmartin:
@agrare can you look please to fix it current sprint? There are errors in QA which should be fixed by this PR

PR #79, opened 2020-04-22
Add the ability to configure ENABLED_SOURCE_TYPES via ENV

lindgrenj6:
https://projects.engineering.redhat.com/browse/TPINVTRY-917

**Depends on:** 
- [x] https://github.com/RedHatInsights/e2e-deploy/pull/1537

In order to only deploy the supported GA collectors on prod, this PR changes the way the application deploys source types, it checks the ENV for which collectors to spin up.

### DIFF
--- a/lib/topological_inventory/orchestrator/source_type.rb
+++ b/lib/topological_inventory/orchestrator/source_type.rb
@@ -3,7 +3,16 @@ require "topological_inventory/orchestrator/api_object"
 module TopologicalInventory
   module Orchestrator
     class SourceType < ApiObject
+      SUPPORTED_TYPES = %w[amazon ansible-tower azure openshift].freeze
       AVAILABILITY_CHECK_SOURCE_TYPES = %w[amazon ansible-tower openshift].freeze
+
+      def method_missing(name, *args, &block)
+        if name.to_s.ends_with?("?")
+          SUPPORTED_TYPES.include?(attributes['name'].to_s)
+        else
+          super
+        end
+      end
 
       def to_s
         attributes['name'] || "Unknown source type: #{attributes.inspect}"
@@ -27,13 +36,6 @@ module TopologicalInventory
 
       def supported_source_type?
         attributes['name'].present? && self[:enabled?]
-      end
-
-      # Methods amazon? / ansible_tower? / azure? / openshift?
-      %w[amazon ansible-tower azure openshift].each do |type|
-        define_method "#{type.sub('-', '_')}?" do
-          attributes['name'].to_s == type
-        end
       end
 
       def supports_availability_check?
--- a/lib/topological_inventory/orchestrator/source_type.rb
+++ b/lib/topological_inventory/orchestrator/source_type.rb
@@ -3,7 +3,6 @@ require "topological_inventory/orchestrator/api_object"
 module TopologicalInventory
   module Orchestrator
     class SourceType < ApiObject
-      SUPPORTED_TYPES = %w[amazon ansible-tower azure openshift].freeze
       AVAILABILITY_CHECK_SOURCE_TYPES = %w[amazon ansible-tower openshift].freeze
 
       def to_s
@@ -27,11 +26,11 @@ module TopologicalInventory
       end
 
       def supported_source_type?
-        attributes['name'].present? && SUPPORTED_TYPES.include?(attributes['name'])
+        attributes['name'].present? && self[:enabled?]
       end
 
       # Methods amazon? / ansible_tower? / azure? / openshift?
-      SUPPORTED_TYPES.each do |type|
+      %w[amazon ansible-tower azure openshift].each do |type|
         define_method "#{type.sub('-', '_')}?" do
           attributes['name'].to_s == type
         end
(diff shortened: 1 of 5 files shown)


lindgrenj6:
cc @slemrmartin @syncrou @gtanzillo

lib/topological_inventory/orchestrator/source_type.rb
```diff
@@ -40,6 +40,12 @@ def supported_source_type?
       def supports_availability_check?
         AVAILABILITY_CHECK_SOURCE_TYPES.include?(attributes['name'])
       end
+
+      private
+
+      def enabled_source_types
+        @enabled_source_types ||= ENV['ENABLED_SOURCE_TYPES']&.split(",") || []
```
slemrmartin, line 47: @lindgrenj6 Please move ENV variable to entrypoint https://github.com/RedHatInsights/topological_inventory-orchestrator/blob/master/bin/topological_inventory-orchestrator. 
With default values equal to current source types

slemrmartin: btw env var can be nil, so `split` will fail in this case

lib/topological_inventory/orchestrator/source_type.rb
```diff
@@ -27,7 +27,7 @@ def collector_definition(collector_image_tag = 'latest')
       end
 
       def supported_source_type?
-        attributes['name'].present? && SUPPORTED_TYPES.include?(attributes['name'])
```
slemrmartin, line 30: you can remove this const's definition

lib/topological_inventory/orchestrator/source_type.rb
lindgrenj6: > Please move ENV variable to entrypoint /bin/topological_inventory-orchestrator@master .
With default values equal to current source types

I'm not 100% sure how you wanted me to do this - but I basically added it as an argument to optimist, then as another argument to the Worker initializer with a default value of the original values.

> btw env var can be nil, so split will fail in this case

Yep, can't risk the NoMethodError, that's why I used the safe operator `&.` which will just return nil if the `ENV` var is nil.

lindgrenj6: I have some spec failures I'm having trouble with, it's just the `targeted_update_spec.rb` file, is there something I'm missing there?

slemrmartin: > that's why I used the safe operator &.

Sure, I was thinking about something else :D

slemrmartin:
@lindgrenj6 Targete update needs to initialize source types too (checked here: https://github.com/RedHatInsights/topological_inventory-orchestrator/blob/f00ff7ca3cf3c3655a6d9003172b1d1eedfdeb62/lib/topological_inventory/orchestrator/targeted_update/destroy_action_rules.rb#L12)

[source-types-targeted-update.zip](https://github.com/RedHatInsights/topological_inventory-orchestrator/files/4527961/source-types-targeted-update.zip)

Try this patch ^^

lindgrenj6:
@slemrmartin The patch worked great! Thank you! Pushed up the changes.

lib/topological_inventory/orchestrator/targeted_update.rb
```diff
@@ -72,6 +75,14 @@ def sync_targets_with_openshift
 
       private
 
+      def enable_supported_source_types
+        @targets.each do |target|
+          if (source_type = target[:source_type])
```
syncrou, line 80: Does this need to be an `==`

lib/topological_inventory/orchestrator/source_type.rb
```diff
@@ -27,11 +26,11 @@ def collector_definition(collector_image_tag = 'latest')
...
+        attributes['name'].present? && self[:enabled?]
       end
 
       # Methods amazon? / ansible_tower? / azure? / openshift?
-      SUPPORTED_TYPES.each do |type|
+      %w[amazon ansible-tower azure openshift].each do |type|
```
gtanzillo, line 33: Does it still make sense to keep the list of supported source type in a constant? I can see us continuing to add to that list as we board additional source types that require collectors and it would be easier to add them to the constant.

lib/topological_inventory/orchestrator/targeted_update.rb
lindgrenj6: re-worked it to remove the inline assignment.

lib/topological_inventory/orchestrator/source_type.rb
slemrmartin: yeah, there we can generate just one method returning true (based on `attributes['name']`). 
Other methods with `?` can be handled by `method_missing`. Better in `is_..._?` format, if it won't take much time.

lib/topological_inventory/orchestrator/source_type.rb
```diff
@@ -6,6 +6,14 @@ class SourceType < ApiObject
       SUPPORTED_TYPES = %w[amazon ansible-tower azure openshift].freeze
       AVAILABILITY_CHECK_SOURCE_TYPES = %w[amazon ansible-tower openshift].freeze
 
+      def method_missing(name, *args, &block)
+        if name.to_s.match?(/is_.*\?/)
```
syncrou, line 10: @lindgrenj6 I like this, but with https://rubocop.readthedocs.io/en/latest/cops_naming/ I suggest going back to `azure?` instead of `is_azure?`

lindgrenj6: Ok, switched back to the old style _but_ it is all handled in `method_missing` now instead of meta-methods.

PR #81, opened 2020-05-07
[WIP] Persister autoscaler

slemrmartin:
Persister autoscaler based on `kafka_consumergroup_group_lag`
http://thanos-query-mnm-ci.5a9f.insights-dev.openshiftapps.com/graph?g0.range_input=1d&g0.max_source_resolution=0s&g0.expr=sum(kafka_consumergroup_group_lag%7Btopic%3D~%27platform.topological-inventory.persister%27%7D)%20by%20(group%2C%20topic)&g0.tab=0

It makes avg from last 300 seconds (step 30s), checking once per 150 seconds.
It scales down if the lag < 25, scales up if the lag is > 75 **[has to be tested]**

Minimum pods = 1, maximum pods = 10.

PR #85, opened 2020-05-28
Prometheus autoscaler

lindgrenj6:
Continuing the work @slemrmartin did on #81, making it more generic. 

\#TODO's inline. 

- [x] depends on: https://github.com/RedHatInsights/e2e-deploy/pull/1754


lindgrenj6:
@slemrmartin @syncrou this is ready for review, the e2e-deploy PR is in.

lib/topological_inventory/orchestrator/metric_scaler.rb
```diff
@@ -1,46 +1,55 @@
...
-      def initialize(logger = nil)
-        require "topological_inventory/orchestrator/metric_scaler/watcher"
+      def initialize(prometheus_hostname, logger = nil)
         @logger = logger || ManageIQ::Loggers::CloudWatch.new
+        @prometheus_hostname = prometheus_hostname
         @cache  = {}
       end
 
       def run
+        logger.info("#{self.class.name}##{__method__} Starting...")
```
syncrou, line 18: With this moved here we only see the log message when the pod starts correct?  Which may be fine, but the original behavior did this every 10 seconds.  

I guess I'm not sure if we need to see this every 10 seconds, so potentially fine. @slemrmartin do you have any thoughts?

lib/topological_inventory/orchestrator/metric_scaler.rb
```diff
@@ -1,46 +1,55 @@
...
-
-        logger.info("#{self.class.name}##{__method__} Complete")
+        dc_names.each                 { |name| @cache[name].scale_to_desired_replicas if @cache[name].scaling_allowed? }
       end
 
       private
 
-      def object_manager
-        @object_manager ||= begin
-          require "topological_inventory/orchestrator/object_manager"
```
syncrou, line 40: I'm thinking this was here for runtime require loading ( potential performance increase? ) I see this was moved to the top of the file.  What were the thoughts moving it to the top vs leaving it here?

lib/topological_inventory/orchestrator/metric_scaler/prometheus_watcher.rb
```diff
@@ -0,0 +1,105 @@
...
+          body = JSON.parse(response.body)
+          raise response.body unless body['status'] == 'success'
+
+          metrics = body.dig('data', 'result').to_a.first
+          if metrics.blank?
+            logger.warn("PrometheusWatcher: Empty result from Prometheus received")
+          end
+          metrics.to_a
+        rescue RestClient::BadRequest => e
+          logger.error("PrometheusWatcher: Bad request: #{e.response.body}")
```
syncrou, line 63: So, if the metrics can't get downloaded, that seems like a major issue.  Should we have more than the logging here?

lib/topological_inventory/orchestrator/metric_scaler/watcher.rb
```diff
@@ -54,28 +72,28 @@ def scale_to_desired_replicas
         private
 
         def metrics
-          @metrics ||= begin
-            require "topological_inventory/orchestrator/fixed_length_array"
```
syncrou, line 58: Same as above.

lib/topological_inventory/orchestrator/metric_scaler/watcher.rb
```diff
@@ -54,28 +72,28 @@ def scale_to_desired_replicas
...
-            TopologicalInventory::Orchestrator::FixedLengthArray.new(60)
-          end
+          @metrics ||= TopologicalInventory::Orchestrator::FixedLengthArray.new(max_metrics_count)
+        end
+
+        def max_metrics_count
+          60
```
syncrou, line 79: Should this be a constant?

lib/topological_inventory/orchestrator/metric_scaler/watcher.rb
```diff
@@ -87,10 +105,7 @@ def pod_ips
         end
 
         def object_manager
-          @object_manager ||= begin
-            require "topological_inventory/orchestrator/object_manager"
```
syncrou, line 91: Same

lib/topological_inventory/orchestrator/metric_scaler/watcher.rb
```diff
@@ -121,7 +136,6 @@ def percent_usage_from_metrics
         end
 
         def scrape_metrics_from_ip(ip)
-          require 'restclient'
```
syncrou, line 124: Same

spec/topological_inventory/orchestrator/metric_scaler/prometheus_watcher_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    it "doesn't allow scaling run when no data collected" do
+      allow(watcher).to receive(:download_metrics).and_return([])
+
+      watcher.start
+      watcher.thread.join
+
+      expect(watcher.scaling_allowed?).to be_falsey
+    end
+
+    it "doesn't allow scaling run when zero data collected" do
```
syncrou, line 68: This is a great spec - which brings up my question above. What if we can't get metrics from Prometheus due to a rescued exception? Should there be an alert for that?  

If there wasn't an exception then there seems to be potential for a lot of lag build up - and we wouldn't know that the Orchestrator couldn't see that and act accordingly.

lib/topological_inventory/orchestrator/metric_scaler.rb
slemrmartin: as you said we don't need this info message each 10 seconds, it doesn't give us too valuable information so this is fine to me

lib/topological_inventory/orchestrator/metric_scaler.rb
slemrmartin: It was my intention, it makes file dependencies more clear when placed on the top of the file. 
Anyway this is loaded only once in both cases so I guess it won't have any impact to performance

lib/topological_inventory/orchestrator/metric_scaler/prometheus_watcher.rb
slemrmartin: Bad request could mean that the PromQL query is wrong in the e2e-deploy template's annotation.
@syncrou what exactly do you mean by having more than logging, Alert? 

In other cases (not HTTP 400) Prometheus is either unavailable or is available, returns HTTP 200, but all values are zero. We can skip scaling in this case but except logging, we will be not able to do more with it

lib/topological_inventory/orchestrator/metric_scaler/prometheus_watcher.rb
```diff
@@ -0,0 +1,105 @@
...
+          uri.path = default_api_path if uri.path.blank?
+          uri.to_s
+        end
+
+        def default_api_path
+          "/api/v1".freeze
+        end
+
+        def promql_query
+          query = "query?query=(#{@query.gsub(" ","")})"
```
slemrmartin, line 99: @lindgrenj6 why do we need this line?

lib/topological_inventory/orchestrator/metric_scaler/prometheus_watcher.rb
```diff
@@ -0,0 +1,105 @@
...
+
+        def configured?
+          @query && @max_replicas && @min_replicas && @target_usage && @scale_threshold
+        end
+
+        def configure
+          @max_replicas        = deployment_config.metadata.annotations["metric_scaler_max_replicas"]&.to_i        # i.e. "5"
+          @min_replicas        = deployment_config.metadata.annotations["metric_scaler_min_replicas"]&.to_i        # i.e. "1"
+          @target_usage        = deployment_config.metadata.annotations["metric_scaler_target_usage_pct"]&.to_f    # i.e. "50"
+          @scale_threshold     = deployment_config.metadata.annotations["metric_scaler_scale_threshold_pct"]&.to_f # i.e. "20"
```
slemrmartin, line 25: `@target_usage` and `@scale_threshold` won't be defined as a percentage value but a size of the Kafka lag, so I would omit the `_pct` suffix

lib/topological_inventory/orchestrator/metric_scaler/watcher.rb
```diff
@@ -54,28 +72,28 @@ def scale_to_desired_replicas
...
         def configure
           logger.info("Fetching configuration for #{deployment_config_name}")
-          @deployment_config   = object_manager.get_deployment_config(deployment_config_name)
           @current_metric_name = deployment_config.metadata.annotations["metric_scaler_current_metric_name"]       # i.e. "topological_inventory_api_puma_busy_threads"
           @max_metric_name     = deployment_config.metadata.annotations["metric_scaler_max_metric_name"]           # i.e. "topological_inventory_api_puma_max_threads"
           @max_replicas        = deployment_config.metadata.annotations["metric_scaler_max_replicas"]&.to_i        # i.e. "5"
           @min_replicas        = deployment_config.metadata.annotations["metric_scaler_min_replicas"]&.to_i        # i.e. "1"
-          @target_usage_pct    = deployment_config.metadata.annotations["metric_scaler_target_usage_pct"]&.to_i    # i.e. "50"
-          @scale_threshold_pct = deployment_config.metadata.annotations["metric_scaler_scale_threshold_pct"]&.to_i # i.e. "20"
+          @target_usage        = deployment_config.metadata.annotations["metric_scaler_target_usage_pct"]&.to_i    # i.e. "50"
```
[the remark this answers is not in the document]
slemrmartin: @syncrou because PrometheusWatcher should scale based on kafka lag (int), not a percentage value

lib/topological_inventory/orchestrator/metric_scaler/watcher.rb
slemrmartin: @syncrou @lindgrenj6 I don't insist in moving it to the top of the file, it doesn't have impact to anything, I'm just curious why do you think it's better than all requires on the top of the file. 
I feel this as something like a rubocop issue (although it obviously isn't in our rubocop rules)

syncrou: @slemrmartin maybe I'm lost in some of the history here, but it appears originally it was being loaded at run time.

If we can load at run time vs all at once, there is potential for performance increases.  

I wasn't 100% saying that we should always load at run time, but if they were originally loaded at runtime ( which isn't bad ) then I didn't see a reason to potentially slow down the performance and simply load them all at once. )

slemrmartin: @syncrou It's just my guess but "require" at the top of the file can slow down the startup time, but speed up runtime, is it correct? 
Anyway it'll be important only when requiring very big files so this is a theoretical discussion

lib/topological_inventory/orchestrator/metric_scaler/prometheus_watcher.rb
syncrou: @slemrmartin @lindgrenj6 - My thought was, if we have no metrics we have no autoscaling, so it seems potentially a Slack alert is warranted.

lindgrenj6: Yeah thats a good point - how about I follow up with that in a different PR since it entails adding prometheus metrics to the metric_scaler itself, and from there creating an alert on it.

syncrou: @lindgrenj6 that works for me.

lindgrenj6: Will follow up here: https://projects.engineering.redhat.com/browse/TPINVTRY-968

PR #87, opened 2020-07-02
Add receptor_node to configmap settings from sources

lindgrenj6:
https://projects.engineering.redhat.com/browse/TPINVTRY-990

This PR adds the `receptor_node` setting to the configmap created by orchestrator (if present). The collector will still have to be updated in order to take advantage of it, but at least with this we can always assume it is there. 

*Depends on:*
- [x] https://github.com/RedHatInsights/e2e-deploy/pull/1916

\# TODO: 
- [x] Add to configmap
- [x] Add configmap setting to digest
- [x] specs
- [x] add setting of queue host/receptor information via env vars

cc @syncrou @slemrmartin

### DIFF
--- a/lib/topological_inventory/orchestrator/config_map.rb
+++ b/lib/topological_inventory/orchestrator/config_map.rb
@@ -263,14 +263,16 @@ module TopologicalInventory
 
       def source_to_yaml(source)
         {
-          :source      => source['uid'],
-          :source_id   => source['id'],
-          :source_name => source['name'],
-          :scheme      => source.endpoint['scheme'],
-          :host        => source.endpoint['host'],
-          :port        => source.endpoint['port'],
-          :path        => source.endpoint['path'],
-          :digest      => source.digest
+          :source         => source['uid'],
+          :source_id      => source['id'],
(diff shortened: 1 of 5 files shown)


slemrmartin:
We need more data for tower:
- `QUEUE_HOST` and `QUEUE_PORT`
- `RECEPTOR_CONTROLLER_[SCHEME|HOST|PORT]`
- `ACCOUNT_NUMBER` (or PSK - for communication with receptor controller)

This is not needed for non-tower deployment templates

lindgrenj6:
> We need more data for tower collector:
> 
>     * `QUEUE_HOST` and `QUEUE_PORT`
> 
>     * `RECEPTOR_CONTROLLER_[SCHEME|HOST|PORT]`
> 
>     * `ACCOUNT_NUMBER` (or PSK - for communication with receptor controller)
> 
> 
> This is not needed for non-tower deployment templates

Ok, I'll add some logic to add those values based off of ENV vars that we pass to orchestrator if the source type has `receptor_node` set.

syncrou:
Added approval - but obviously still needs the e2edeploy PR.

PR #88, opened 2020-07-20
Update DeploymentConfig resource to pull image definition from running operations workers

lindgrenj6:
Updating orchestrator to be a bit smarter, go off of the currently running operations pods for collector definitions rather than just using what orchestrator is running. 

\# TODO: 
- [x] Pull collector image from running operations pods
- [x] specs for deployment_config.rb
- [x] store currently-running tag (git sha) in the ConfigMap resource
- [x] redeploy worker if tag changes
- [x] ~specs for update-process~ included in the configmap digest logic already

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -8,7 +8,7 @@ require "topological_inventory-orchestrator"
 require "irb"
 
 def default_worker
-  TopologicalInventory::Orchestrator::Worker.new(collector_image_tag: ENV["IMAGE_TAG"], sources_api: ENV["SOURCES_API"], topology_api: ENV["TOPOLOGICAL_INVENTORY_API"])
+  TopologicalInventory::Orchestrator::Worker.new(sources_api: ENV["SOURCES_API"], topology_api: ENV["TOPOLOGICAL_INVENTORY_API"])
 end
 
 IRB.start
--- a/lib/topological_inventory/orchestrator/config_map.rb
+++ b/lib/topological_inventory/orchestrator/config_map.rb
@@ -68,7 +68,7 @@ module TopologicalInventory
           source = sources_by_digest[digest]
           if source.nil?
             # This source is not in API, will be deleted from openshift
-            source = Source.new({}, nil, source_type, nil, :from_sources_api => false)
+            source = Source.new({}, nil, source_type, :from_sources_api => false)
             source.digest = digest
             sources_by_digest[digest] = source
             logger.debug("Assoc Map (#{uid}) -> Source (digest #{digest}) not found")
@@ -272,6 +272,7 @@ module TopologicalInventory
           :path           => source.endpoint['path'],
           :receptor_node  => source.endpoint['receptor_node'],
           :account_number => source.tenant,
+          :image_tag_sha  => source.source_type['collector_image'].split(":").last,
           :digest         => source.digest
         }
       end
--- a/lib/topological_inventory/orchestrator/object_manager.rb
+++ b/lib/topological_inventory/orchestrator/object_manager.rb
@@ -32,8 +32,8 @@ module TopologicalInventory
         )
       end
 
-      def create_deployment_config(name, image_namespace, image)
-        definition = deployment_config_definition(name, image_namespace, image)
+      def create_deployment_config(name, type)
+        definition = deployment_config_definition(name, type)
         yield(definition) if block_given?
         check_deployment_config_quota(definition)
         connection.create_deployment_config(definition)
@@ -113,6 +113,15 @@ module TopologicalInventory
       def delete_replication_controller(name)
         kube_connection.delete_replication_controller(name, my_namespace)
       rescue Kubeclient::ResourceNotFoundError
+      end
+
+      def get_collector_image(name)
+        pod = kube_connection.get_pods(
+          :namespace      => my_namespace,
+          :label_selector => "name=topological-inventory-#{name}-operations"
+        ).first
+
+        pod&.spec&.containers&.first&.image
       end
 
       private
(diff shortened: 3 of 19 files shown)


lib/topological_inventory/orchestrator/object_manager.rb
```diff
@@ -115,6 +115,15 @@ def delete_replication_controller(name)
       rescue Kubeclient::ResourceNotFoundError
       end
 
+      def get_collector_image(name)
+        pod = connection.get_pods(
+          :namespace => my_namespace,
+          :label_selector => "name=topological-inventory-#{name}-operations"
```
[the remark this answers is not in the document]
lindgrenj6: I hooked up my local to the OCP4 API and there is not a `get_deployments` method generated like I thought there would be, the kubeclient gem metaprograms the methods based on what it discovers. Maybe it's different when running in OCP itself - but I can't exec in to figure it out for now. 

If that method does happen to exist though it will be as easy as changing it out and changing the `.spec.containers.first.image` -> `.spec.template.spec.containers.first.image`.

lindgrenj6:
@syncrou @slemrmartin This is ready for review. I just have to write a few specs for making sure the full-refresh will redeploy everything.

bin/console
```diff
@@ -8,7 +8,7 @@ require "topological_inventory-orchestrator"
 require "irb"
 
 def default_worker
-  TopologicalInventory::Orchestrator::Worker.new(collector_image_tag: ENV["IMAGE_TAG"], sources_api: ENV["SOURCES_API"], topology_api: ENV["TOPOLOGICAL_INVENTORY_API"])
+  TopologicalInventory::Orchestrator::Worker.new(sources_api: ENV["SOURCES_API"], topology_api: ENV["TOPOLOGICAL_INVENTORY_API"])
```
syncrou, line 11: Less arguments is always good. 🌮

lib/topological_inventory/orchestrator/config_map.rb
```diff
@@ -272,6 +272,7 @@ def source_to_yaml(source)
           :path           => source.endpoint['path'],
           :receptor_node  => source.endpoint['receptor_node'],
           :account_number => source.tenant,
+          :image          => source.source_type['collector_image'],
```
syncrou, line 275: Should this naming designate the `sha` - I believe 'collector_image' is detailed, but to help understand that key expects a sha tag vs just 'any' tag could potentially make the code more readable.

lib/topological_inventory/orchestrator/object_manager.rb
```diff
@@ -115,6 +115,15 @@ def delete_replication_controller(name)
       rescue Kubeclient::ResourceNotFoundError
       end
 
+      def get_collector_image(name)
+        pod = connection.get_pods(
+          :namespace      => my_namespace,
+          :label_selector => "name=topological-inventory-#{name}-operations"
+        ).first
+
+        pod&.spec&.container&.first&.image
```
syncrou, line 124: What if this returns `nil` because we have no collector pods running? 

It seems like a potential race condition?

lindgrenj6: Well, it goes off of the running `operations` pod not the currently running collector(s), but nonetheless it is a race condition if the orchestrator were to come up before the operations pod was ready. It would eventually catch up though since the operations pods have to be running in order for any collectors to be of use. 

Basically this change makes the orchestrator dependent on the operations pods. I do handle if the `collector_image` is nil - we skip that source. So it would catch up eventually and spin up the collector.

lib/topological_inventory/orchestrator/config_map.rb
lindgrenj6: I'm down for that change - but can I do it in a followup? I would have to regenerate the test-data digests which is pretty time consuming.

lindgrenj6: Updated it to `:image_tag_sha`

lib/topological_inventory/orchestrator/object_manager.rb
```diff
@@ -115,6 +115,15 @@ def delete_replication_controller(name)
       rescue Kubeclient::ResourceNotFoundError
       end
 
+      def get_collector_image(name)
+        pod = connection.get_pods(
```
syncrou, line 119: So this works as is even with the `kube_connection` vs. `ocp_connection` - Or was that issue else where?

lindgrenj6: Just pushed the update for `kube_connection`, I hadn't pushed it yet, sorry!

spec/topological_inventory/orchestrator/test_models/kube_client.rb
```diff
@@ -82,6 +84,10 @@ def delete_secret(name, _namespace)
           secrets.delete(name)
         end
 
+        def get_pods(namespace:, label_selector:)
```
gmcculloug, line 87: @lindgrenj6 Could we just throw a quick `_namespace` here to keep rubocop happy.  Seems to be there pattern above so maybe need to think about a followup PR to adjust for the fact that it is not used anymore.

It's ok to ignore the `find_by` warning for now as that is defined as a private method here and is not a call to the  Rails dynamic finder.

lindgrenj6: Done, I'm curious if we can inform rubocop that we don't even have rails here. He isn't quite sure yet haha.

gmcculloug: We could look into disabling that cop on the file, see https://docs.rubocop.org/rubocop/configuration.html#disabling-cops-within-source-code.

We can do that in a followup PR.

PR #90, opened 2020-07-23
Force Sync of Collector Images if DeploymentConfig was not re-created

lindgrenj6:
So after testing a bit today I found if the configmap doesn't get completely deleted it won't create a new DeploymentConfig with the new image, this alleviates that by running a "sync" between the config_maps/deployment_configs on their images. 

\# TODO: 
- [x] specs

cc @syncrou @gmcculloug

### DIFF
--- a/lib/topological_inventory/orchestrator/deployment_config.rb
+++ b/lib/topological_inventory/orchestrator/deployment_config.rb
@@ -72,6 +72,29 @@ module TopologicalInventory
         logger.info("[OK] Deleted DeploymentConfig #{self}")
       end
 
+      def update_image(new_image)
+        logger.info("Updating DeploymentConfig Image #{self}")
+
+        # This is ugly, but its a JSON patch we're sending.
+        patch = {
+          :spec => {
+            :template => {
+              :spec => {
+                :containers => [
+                  :name  => name,
+                  :image => new_image
+                ]
+              }
+            }
+          }
+        }
+
+        object_manager.update_deployment_config(name, patch)
+        openshift_object(:reload => true)
+
+        logger.info("[OK] Updated DeploymentConfig Image #{self}")
+      end
+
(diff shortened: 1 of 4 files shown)


lib/topological_inventory/orchestrator/deployment_config.rb
```diff
@@ -72,6 +72,28 @@ def delete_in_openshift
...
+      def update_image(image)
+        logger.info("Updating DeploymentConfig Image #{self}")
+
+        # This is ugly, but its a JSON patch we're sending.
+        patch = {
+          :spec => {
+            :template => {
+              :spec => {
+                :containers => [
+                  :name  => openshift_object.spec.template.spec.containers.first.image,
```
syncrou, line 84: Could we just use the `image` method here too?

lindgrenj6: I messed up there, it was late. I meant to pull the name of the container. Also updated the name of the var from `image` to `new_image`.

spec/worker_black_box_spec.rb
```diff
@@ -277,5 +277,70 @@
...
+        sources_from_api = [topological_sources[:openshift]['1'],
+                            topological_sources[:amazon]['4'],
+                            topological_sources[:amazon]['6'], # new
+                            topological_sources[:azure]['7']]  # new
+
+        # :application_types request cached
+        stub_api_init(:applications      => applications_response,
+                      :source_types      => source_types_response,
+                      :tenants           => tenants_response,
+                      :topology_sources  => list(sources_from_api))
```
gmcculloug, line 333: There are two spaces before the hash rocket which is causing the rubocop error.  Just need to remove one and align the lines above.

lindgrenj6:
@syncrou this is ready to go!

PR #96, opened 2020-07-24
Move Scheduled Sync Event to an ENV var rather than a config-file setting

lindgrenj6:
Related: https://github.com/RedHatInsights/e2e-deploy/pull/2022

This will still default to 1 hour, but we can override this a bit easier per-env.


lindgrenj6:
cc @syncrou @slemrmartin

PR #98, opened 2020-07-29
Update Source#digest to account for receptor-enabled sources

lindgrenj6:
We weren't accounting for receptor-enabled sources in the `Source` file yet - thus the digest wasn't being computed so the collectors weren't getting spun up. This adds a check for `receptor_node` and handles it appropriately.

### DIFF
--- a/lib/topological_inventory/orchestrator/source.rb
+++ b/lib/topological_inventory/orchestrator/source.rb
@@ -67,7 +67,10 @@ module TopologicalInventory
 
       def digest(reload: false)
         return @digest if @digest.present? && !reload
-        return nil if attributes.nil? || endpoint.nil? || authentication.nil? || credentials.nil?
+        return nil if attributes.nil? || endpoint.nil?
+
+        # If we are not going through a receptor node
+        return nil if endpoint["receptor_node"].blank? && (authentication.nil? || credentials.nil?)
 
         @digest = compute_digest(digest_values)
       end
@@ -88,8 +91,8 @@ module TopologicalInventory
           "source_id"       => attributes["id"],
           "source_uid"      => attributes["uid"],
           "secret"          => {
-            "password" => credentials["password"],
-            "username" => credentials["username"],
+            "password" => credentials.try(:[], "password"),
+            "username" => credentials.try(:[], "username"),
           }
         }
         # Azure has extra parameter "tenant_id"
--- a/spec/source_spec.rb
+++ b/spec/source_spec.rb
@@ -1,9 +1,10 @@
 describe TopologicalInventory::Orchestrator::Source do
   include MockData
+  let(:source_type) { double(:[] => "quay.io/ansible:123abc", :azure? => false) }
 
   before do
     @source = described_class.new(sources_data[:openshift]["1"], nil,
-                                  double, :from_sources_api => true)
+                                  source_type, :from_sources_api => true)
   end
 
   context "#add_to_openshift" do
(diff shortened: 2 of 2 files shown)


lindgrenj6:
cc @syncrou @slemrmartin

lib/topological_inventory/orchestrator/source.rb
```diff
@@ -55,10 +57,15 @@ def load_credentials(api)
         self.endpoint = api.get_endpoint(attributes['id'], tenant)
         return if endpoint.nil?
 
-        self.authentication = api.get_authentication(endpoint['id'], tenant)
-        return if authentication.nil?
+        if endpoint["receptor_node"].present?
+          self.authentication = AUTH_NOT_NECESSARY
```
mkanoor, line 61: @lindgrenj6 Does it make sense to call it DELEGATE_AUTHENTICATION_TO_RECEPTOR instead of AUTH_NOT_NECESSARY? Since the receptor is the one that is storing the credentials or tokens which will be used when connecting to the Source.

mkanoor: Another thing I noticed in the sources database is authtype: "receptor_node"

```
#<Authentication id: 10820, resource_type: "Endpoint", resource_id: 10898, name: nil, authtype: "receptor_node", 
username: nil, password: nil, status: nil, status_details: nil, tenant_id: 2, extra: nil, availability_status: nil, 
availability_status_error: nil, last_checked_at: nil, last_available_at: nil>
```

mkanoor: https://github.com/RedHatInsights/sources-api/blob/69289f0f834f1b37f4d0d8c96ab5f09101967f7e/db/seeds/application_types.yml#L55

mkanoor: Should we be adding another authtype here to be in sync with satellite
https://github.com/RedHatInsights/sources-api/blob/69289f0f834f1b37f4d0d8c96ab5f09101967f7e/db/seeds/application_types.yml#L10

lindgrenj6: We probably will make orchestrator smarter in the future, but right now we're just trying to get the current iteration working with receptor sources. Right now Orchestrator doesn't even look at AuthTypes, we just grab authentications. 

I updated the logic to just check in the `digest` method since that was all that was breaking before, now we don't need a constant either.

spec/source_spec.rb
```diff
@@ -48,21 +49,41 @@
...
-      endpoint, authentication, credentials = double, double, double
-      allow(endpoint).to receive(:[])
-      allow(authentication).to receive(:[])
+    context "when there is a receptor node" do
+      it "still computes the digest even without auth/creds" do
+        endpoint = {"receptor_node" => "asdf"}
+        api = double("api")
+
+        allow(api).to receive_messages(:get_endpoint       => endpoint,
+                                       :get_authentication => nil,
```
mkanoor, line 58: @lindgrenj6 do you want to set the get_authentication to return 'receptor_node'? Since the authtype is set to receptor_node.

lindgrenj6: I mean I could - but we dont do anything with that data. The _only_ piece we care about is the `receptor_node` field on the `Endpoint`, everything else is optional (as far as orchestrator is concerned). As long as we have the `receptor_node`, we can spin up the collector. 

@syncrou maybe I'm missing something here - we don't use the authType anywhere right? I'm not aware that we do.

mkanoor:
@lindgrenj6 @syncrou 
authtype of receptor_node is a valid type. It is listed here
https://github.com/RedHatInsights/sources-api/blob/69289f0f834f1b37f4d0d8c96ab5f09101967f7e/db/seeds/application_types.yml#L57

I think it was introduced for Satellite which till now has been the only source with receptor support.

But Satellite never needed a collector or operations worker. So it was never excercised.

Now that we have a source that needs a collector and operations worker we should try and use that authtype.


Also I am thinking that we need to add a new entry here for Ansible Tower in this PR

https://github.com/RedHatInsights/sources-api/blob/69289f0f834f1b37f4d0d8c96ab5f09101967f7e/db/seeds/application_types.yml#L10

This would allows us to have Towers which can be accessed via userid/password and also via receptor.

PR #100, opened 2020-07-31
Change SCHEDULED_SYNC_HOURS from integer to float

lindgrenj6:
This would truncate if `0.5` was used for example. Found when making sure the code was on CI.


lindgrenj6:
cc @syncrou catching this before it happens.

PR #101, opened 2020-08-05
Add ENV vars so we can modify collector resource definitions on the fly

lindgrenj6:
This way we do not have to ship a code change in order to tweak the cpu/memory allocation.

e2e-deploy PR: https://github.com/RedHatInsights/e2e-deploy/pull/2073

### DIFF
--- a/lib/topological_inventory/orchestrator/object_manager.rb
+++ b/lib/topological_inventory/orchestrator/object_manager.rb
@@ -269,12 +269,12 @@ module TopologicalInventory
                   :image     => image,
                   :resources => {
                     :limits   => {
-                      :cpu    => "50m",
-                      :memory => "400Mi"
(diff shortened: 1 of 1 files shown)


lindgrenj6:
@slemrmartin I did add them (if present) to the digest - so if we change them it should invalidate the deployment config and trigger a re-create :grin:

slemrmartin:
@lindgrenj6 digests change will remove all the sources and add them to new deployment configs. 
But it can also move them to any existing deployment configs if there'll be free slot.

I'm wondering what happens with the existing deployment configs, some sources can be removed and some can be added at the same time, so theoretically all the deployment configs may not be recreated

lindgrenj6:
@slemrmartin Ahh good point. Yeah I can potentially add code to sync the resources (like we do with image tags), but that will be more work. Maybe eventually.

PR #102, opened 2020-08-06
Fix prometheus response coming back as a sub-sub-array

lindgrenj6:


lib/topological_inventory/orchestrator/metric_scaler/prometheus_watcher.rb
```diff
@@ -33,7 +33,7 @@ def start
             until finished?
               new_metrics = download_metrics
               if new_metrics.present?
-                new_metrics.each do |time_value_pair|
+                new_metrics[1][1].each do |time_value_pair|
```
syncrou, line 36: Did this change on OSD3? OSD4? Both?  Was there a version change?

lindgrenj6: Not sure - it may have been I was testing locally and my local prometheus was returning differently formatted, either that or something changed recently with mnm.

lindgrenj6: I don't like that I missed this, but I tested it this time on CI to see if the formatting was right, and it is now.

syncrou: Do we know that it is the same on OSD4?

lindgrenj6: Asking kyle now, there is only one prometheus for OCP4 prod (https://prometheus.crcp01ue1.devshift.net/) and it requires auth. I'm hoping theres an internal prometheus we can hit without auth, otherwise this will not work on OCP4.

lindgrenj6: This should work on OCP4, we have an authless prometheus here: `prometheus-app-sre.openshift-customer-monitoring.svc`

PR #103, opened 2020-08-12
Fix scaling to go off of pods under deployment rather than counting pod IPs

lindgrenj6:


lib/topological_inventory/orchestrator/metric_scaler/prometheus_watcher.rb
```diff
@@ -75,12 +75,31 @@ def download_metrics
...
+
           @scaling_allowed.value = false
         end
 
         private
 
+        def worker_pods
+          object_manager
+            .get_pods
+            .select { |pod| pod.metadata.annotations["openshift.io/deployment-config.name"] == deployment_config_name }
```
syncrou, line 99: Would this have to change later when we start using deployments instead of deployment-configs?

lindgrenj6: > I'm not seeing, where we remove the code that was counting ips instead of Running pods?

The change is in the superclass, we were just calling `super` before and that was failing. 


> Would this have to change later when we start using deployments instead of deployment-configs?

Yes - this will get changed when we move orchestrator to use Deployments

PR #105, opened 2020-08-21
Sync the running collector(s) resources to what we have configured in the ENV

lindgrenj6:
This way if we need to bump resources on the collectors, it is as easy as bumping them in the env settings in the orchestrator, it will update all running collectors.


lindgrenj6:
@dippy-bot add_reviewer syncrou

PR #106, opened 2020-09-01
[WIP] Migrate Scheduler from individual repo into Orchestrator

lindgrenj6:
Just toying with this idea. That way we wouldn't have to have a whole different repository/buildconfig for this image since it's only 2 files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /.bundle/
 /Gemfile.lock
+/config/database.yml
 
 # generated rubocop file
 /.rubocop-https*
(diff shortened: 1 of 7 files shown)


slemrmartin:
@lindgrenj6 is it 1:1 copy?

lindgrenj6:
@slemrmartin yes, only difference is I changed the logging class to use the orchestrator logging class. I haven't updated it here to construct the `DATABASE_URL` env var from parameters yet, I assume you're going to do that in the scheduler repo. This was just more of an experiment to see how much work it would be if we run into issues building and pushing the new image to quay. 

I did find one bug when running it locally - there are some Tasks in the DB with no `forwardable_headers`, so we should probably account for that. 

I'm ok with using the other repo as well if we're going to be expanding this out more i.e. all of our jobs will live in that scheduler repo etc, I just worry about the build/push not working for the new AppSRE environments.

lindgrenj6:
Closing this since the scheduler build seems to have went fine.

PR #109, opened 2020-09-25
Retry Collector Deployments if the availableReplicas are 0

lindgrenj6:
Currently seeing an issue on stage where a Collector failed to fire up (we were over the quota), this change makes the orchestrator delete the old dc and retry creating it.

\# TODO:
- [x] specs

### DIFF
--- a/lib/topological_inventory/orchestrator/deployment_config.rb
+++ b/lib/topological_inventory/orchestrator/deployment_config.rb
@@ -8,6 +8,7 @@ module TopologicalInventory
       LABEL_COMMON = "tp-inventory/collector".freeze
       LABEL_DIGEST = "topological-inventory/collector_digest".freeze # single-source DCs
       LABEL_UNIQUE = "tp-inventory/config-uid".freeze
+      RECREATE_RETRY_COUNT = 3
 
       attr_accessor :config_map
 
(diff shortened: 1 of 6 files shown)


lib/topological_inventory/orchestrator/object_manager.rb
```diff
@@ -31,6 +31,8 @@ def delete_pod(name)
 
       def get_deployment_config(name)
         connection.get_deployment_config(name, my_namespace)
+      rescue KubeException
+        nil
```
syncrou, line 35: Should we include a logged messge here? Just asking because it stuck out to me.

lib/topological_inventory/orchestrator/deployment_config.rb
```diff
@@ -72,6 +72,24 @@ def delete_in_openshift
         logger.info("[OK] Deleted DeploymentConfig #{self}")
       end
 
+      def recreate_in_openshift
+        delete_in_openshift
+        retry_count = 0
+
+        until retry_count == 3
```
syncrou, line 79: Probably overkill, but just asking. Is there any reason this should either be an env var? ie will 3 be too many?  Not enough?

Or at the very least should potentially be a constant for better exposure?

lindgrenj6: Yeah - I like the idea of changing it to a constant. 30 seconds should be enough time for the resource to hopefully disappear, though I'm not quite sure we'd need it as an ENV var. I'm not sure we'd ever have to increase it.

lib/topological_inventory/orchestrator/object_manager.rb
lindgrenj6: Sure, only reason I didn't is we aren't doing it anywhere else. I'll add it to the rest of the rescues in the future!

syncrou (approved):
LGTM 👍

syncrou:
@slemrmartin can you take a final look?

slemrmartin (approved):
That's great! :+1:

PR #111, opened 2020-10-05
Add log statements to Object Manager

lindgrenj6:
Bugging out on stage because the `include Logging` statement wasn't at the top, went ahead and added some more logging as well in the exception blocks.


syncrou (approved):
LGTM 👍

PR #112, opened 2020-10-19
Orchestrator: Custom metrics

slemrmartin:
**Issue** https://github.com/RedHatInsights/topological_inventory-api/issues/320

Adding prometheus exporter and metrics to the orchestrator.
* event_counter (by event type)
* errors_counter (by error type)
* config maps gauge (by source type)
* deployment configs gauge (by source type)
* secrets gauge (by source type

- [x] TODO: specs

---

[RHCLOUD-9942](https://issues.redhat.com/browse/RHCLOUD-9942)

### DIFF
--- a/bin/metrics_scaler
+++ b/bin/metrics_scaler
@@ -11,10 +11,10 @@ args = Optimist.options do
   opt :prometheus_port, "Port of the prometheus metrics server", :type => :integer, :default => ENV['PROMETHEUS_PORT']&.to_i
 end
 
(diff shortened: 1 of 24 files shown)


lindgrenj6 (approved):
So the good news is MetricScaler is very much going to go away since it doesn't work in OCP4, so I'd almost be for pulling out those metrics (metric_scaler is not running in stage or production since it doesn't work). 

Other than that I think we're good to go here. Let me know if you want to remove those Metric Scaler specs - otherwise I'll merge.

PR #113, opened 2020-10-26
Add LOG_LEVEL ENV var to collectors

lindgrenj6:
So we have the `LOG_LEVEL` settings already in providers common: https://github.com/RedHatInsights/topological_inventory-providers-common/blob/792b61f6cb7b3b6af6e4b180847c36a51dbbb83f/lib/topological_inventory/providers/common/logging.rb#L41

We just need to pass it through. I will have someone from AppSRE edit the currently running collectors - but this will fix it for the future.

### DIFF
--- a/lib/topological_inventory/orchestrator/deployment_config.rb
+++ b/lib/topological_inventory/orchestrator/deployment_config.rb
@@ -166,6 +166,7 @@ module TopologicalInventory
 
       def container_env_values
         [
+          {:name => "LOG_LEVEL", :value => ENV['COLLECTOR_LOG_LEVEL'] || "WARN"},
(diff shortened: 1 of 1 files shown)


lindgrenj6:
Ran this on CI and checked its working:

`/tmp/env_patch.json`
```json
{
  "spec": {
    "template": {
      "spec": {
        "containers": [{
          "name": "NAMEY",
          "env": [
            {
              "name": "LOG_LEVEL",
              "value": "WARN"
            }
          ]
        }]
      }
    }
  }
}
```

```bash
for i in $(oc get dc | awk '/collector/ {print $1}'); do 
    sed -i "s/NAMEY/$i/g" /tmp/env_patch.json 
    oc patch dc/$i --patch "$(cat /tmp/env_patch.json)" 
    sed -i "s/$i/NAMEY/g" /tmp/env_patch.json 
    oc rollout latest dc/$i
done
```

slemrmartin (approved):
:+1:

PR #116, opened 2020-11-03
Add Liveness Probe checking Sources, Topology, and k8s APIs

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-10440

Adding a liveness probe for orchestrator. It goes out and checks Sources and Topology for connectivity, then hits the OCP API to see if our auth is working.

### DIFF
--- a/lib/topological_inventory/orchestrator/event_manager.rb
+++ b/lib/topological_inventory/orchestrator/event_manager.rb
@@ -24,6 +24,7 @@ module TopologicalInventory
       def run!
         Thread.new { event_listener }
         Thread.new { scheduler }
+        Thread.new { health_checker }
 
         processor
       end
@@ -101,6 +102,16 @@ module TopologicalInventory
       rescue => e
         logger.error("#{e.message}\n#{e.backtrace.join('\n')}")
         metrics&.record_error(:event_manager)
+      end
+
+      # Check the (k8s|topo|sources) apis to make sure we're healthy
+      def health_checker
+        sources_api = URI.parse(worker.api.sources_api)
+        topology_api = URI.parse(worker.api.topology_api)
+        k8s = worker.send(:object_manager)
+        interval = worker.send(:health_check_interval)
+
+        HealthCheck.new(sources_api, topology_api, k8s, interval).run
       end
 
       def persist_ref
--- a/lib/topological_inventory/orchestrator/health_check.rb
+++ b/lib/topological_inventory/orchestrator/health_check.rb
@@ -1,0 +1,61 @@
+module TopologicalInventory
+  module Orchestrator
+    class HealthCheck
+      include Logging
+
+      attr_accessor :sources_api, :topology_api, :k8s, :interval
+
+      def initialize(sources_api, topology_api, k8s, interval)
+        @sources_api = sources_api
+        @topology_api = topology_api
+        @k8s = k8s
+        @interval = interval
+      end
+
+      def run
+        loop do
+          checks
+
+          sleep interval
+        end
+      end
+
+      def checks
+        errors = []
+
+        errors << check_url(topology_api)
+        errors << check_url(sources_api)
+        errors << check_k8s(k8s)
+
+        if errors.any?
+          FileUtils.rm_f("/tmp/healthy")
+        else
+          FileUtils.touch("/tmp/healthy")
+        end
+      end
+
+      private
+
+      def check_url(uri)
+        logger.info("[HealthCheck] Checking URL: #{uri}")
+
+        resp = Net::HTTP.get_response(uri.host, "/health", uri.port)
+        # Checking if the response code is 400 or larger
+        # this is the exact same behavior as OCP's http livenessProbe
+        raise if resp.code.to_i >= 400
+      rescue
+        logger.warn("[HealthCheck] Failed URL: #{uri}")
+        :error
+      end
+
+      def check_k8s(k8s)
+        logger.info("[HealthCheck] Checking k8s: #{k8s.send(:connection).api_endpoint}")
+
+        raise if k8s.check_api_status.nil?
+      rescue
+        logger.warn("[HealthCheck] k8s failing")
+        :error
+      end
+    end
+  end
+end
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,8 @@ RUN dnf -y --disableplugin=subscription-manager module enable ruby:2.5 && \
       gcc-c++ make redhat-rpm-config \
       # For git based gems
       git \
+      # For checking service status
+      nmap-ncat \
       && \
     dnf --disableplugin=subscription-manager clean all
 
--- a/bin/topological_inventory-orchestrator
+++ b/bin/topological_inventory-orchestrator
@@ -16,6 +16,8 @@ def parse_args
     opt :config, "Configuration YAML file name", :type => :string, :default => ENV["CONFIG"] || 'default'
     opt :source_types, "The Source types to spin up collectors/operations pods for", :type => :string,
         :default => ENV['ENABLED_SOURCE_TYPES']
+    opt :health_check_interval, "The interval to run health-checks against sources/topo/k8s APIs", :type => :integer,
+        :default => (ENV["HEALTH_CHECK_INTERVAL"] || 60).to_i
   end
 
   opts
@@ -38,6 +40,7 @@ w = TopologicalInventory::Orchestrator::Worker.new(
   :topology_api => args[:topology_api],
   :config_name  => args[:config],
   :source_types => args[:source_types]&.split(","),
-  :metrics      => metrics
+  :metrics      => metrics,
+  :health_check_interval => args[:health_check_interval]
 )
 w.run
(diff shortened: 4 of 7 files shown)


lib/topological_inventory/orchestrator/event_manager.rb
```diff
@@ -103,6 +104,48 @@ def process_events(events)
...
+          errors << check_url(sources)
+          errors << check_k8s(k8s)
+
+          if errors.any?
+            FileUtils.rm_f("/tmp/healthy")
+          else
+            FileUtils.touch("/tmp/healthy")
+          end
+
+          sleep 60
```
syncrou, line 126: Should this be managed by an env variable? If not is there a way for the deployment configuration to fall out of sync with this integer for sleep?

lindgrenj6: Good point. I'll update the PR to have a template for the interval, that way one update will update them both and they're always in sync.

syncrou (commented):
So, is there any possibility of specs? But I like the new `HealthCheck` class a lot.

lib/topological_inventory/orchestrator/health_checker.rb
```diff
@@ -0,0 +1,52 @@
+module TopologicalInventory
+  module Orchestrator
+    class HealthChecker
```
slemrmartin, line 3: I suggest to you `HealthCheck` instead of `HealthChecker` to be consistent with operations repos

lib/topological_inventory/orchestrator/health_checker.rb
```diff
@@ -0,0 +1,52 @@
...
+          errors << check_url(sources)
+          errors << check_k8s(k8s)
+
+          if errors.any?
+            FileUtils.rm_f("/tmp/healthy")
+          else
+            FileUtils.touch("/tmp/healthy")
+          end
+
+          sleep ENV['HEALTH_CHECK_INTERVAL']&.to_i || 60
```
slemrmartin, line 28: can we move it either to top of the class or as an input arg? I would like to have envs more visible

lib/topological_inventory/orchestrator/health_checker.rb
```diff
@@ -0,0 +1,52 @@
...
+        end
+      end
+
+      def check_url(url)
+        logger.info("[HealthCheck] Checking URL: #{url}")
+
+        `ncat -z -w 1 #{url.host} #{url.port}`
```
slemrmartin, line 35: why `ncat` instead of standard ruby calls?

lib/topological_inventory/orchestrator/health_checker.rb
```diff
@@ -0,0 +1,52 @@
...
+        if $?.exitstatus != 0
+          logger.warn("[HealthCheck] Failed URL: #{url}")
+          :error
+        end
+      end
+
+      def check_k8s(k8s)
+        logger.info("[HealthCheck] Checking k8s: #{k8s.send(:connection).api_endpoint}")
```
slemrmartin, line 43: could this raise an exception?

lib/topological_inventory/orchestrator/health_checker.rb
slemrmartin: this will result to 0, not to 60 in case of missing env

lib/topological_inventory/orchestrator/health_checker.rb
lindgrenj6: Updated to use the stdlib `Net::HTTP` and hit the `/health` endpoints on the services.

lib/topological_inventory/orchestrator/health_checker.rb
lindgrenj6: Yep good point, made the error handling exception-based.

lib/topological_inventory/orchestrator/health_check.rb
```diff
@@ -0,0 +1,62 @@
...
+      end
+
+      def checks
+        errors = []
+
+        errors << check_url(topology)
+        errors << check_url(sources)
+        errors << check_k8s(k8s)
+
+        if errors.any?
```
slemrmartin, line 29: I'm wondering if it isn't too strict. 
On one side, the orchestrator is live is at least one check is true.
On the other side, the orchestrator can't work properly if it doesn't have access to all - but we're not as strict in other services

lindgrenj6: Yeah, I'm of the opinion that if we can't hit one of the APIs then orchestrator isn't going to be functioning correctly. 

In order for the pod to restart as well it would have to fail 3 times in a row (which means for 3 minutes straight). 

I think it's reasonable, but would like to hear everyone's opinion cc @syncrou @roliveri 

---

We could potentially make it be 2+ required to mark it un-healthy etc too, or require all to go down.

lindgrenj6:
@roliveri @syncrou thoughts?

lindgrenj6:
@roliveri @slemrmartin could I get some :eyes: on this? Or at least some discussion on what it takes to be "down".

lib/topological_inventory/orchestrator/health_check.rb
roliveri: Sorry I missed the email notifications of your request.

If access to all is required for proper function, then I would think it isn't healthy if any cannot be reached.
Does the fact that some of the failures may be transient in nature come into play?

lindgrenj6: Yes - we account for that in the OCP deployment. It checks once a minute and if it fails for 3 minutes in a row _then_ it will be restarted. That way if it fails once it is just a blip.

lib/topological_inventory/orchestrator/health_check.rb
```diff
@@ -0,0 +1,62 @@
...
+        logger.info("[HealthCheck] Checking k8s: #{k8s.send(:connection).api_endpoint}")
+
+        raise if k8s.check_api_status.nil?
+      rescue
+        logger.warn("[HealthCheck] k8s failing")
+        :error
+      end
+
+      def configured_seconds
+        @configured_seconds ||= ENV['HEALTH_CHECK_INTERVAL'] ? ENV['HEALTH_CHECK_INTERVAL'].to_i : 60
```
slemrmartin, line 58: @lindgrenj6 can you make it more visible? At least add comment to `/bin/topological_inventory-orchestrator`

lib/topological_inventory/orchestrator/health_check.rb
slemrmartin: So let's keep it as is

lib/topological_inventory/orchestrator/health_check.rb
```diff
@@ -0,0 +1,62 @@
+module TopologicalInventory
+  module Orchestrator
+    class HealthCheck
+      include Logging
+
+      attr_accessor :sources, :topology, :k8s
+
+      def initialize(sources, topology, k8s)
+        @sources = sources
```
slemrmartin, line 9: just a small suggestion, can you rename it to @sources_api/@topology_api to be consistent with the rest of code?

lib/topological_inventory/orchestrator/health_check.rb
```diff
@@ -0,0 +1,62 @@
...
+        end
+      end
+
+      private
+
+      def check_url(uri)
+        logger.info("[HealthCheck] Checking URL: #{uri}")
+
+        resp = Net::HTTP.get_response(uri.host, "/health", uri.port)
+        raise if resp.code.to_i != 200
```
slemrmartin, line 42: is it possible it'll be redirected (status 30x)? i.e. when new api version will be released

lib/topological_inventory/orchestrator/health_check.rb
slemrmartin: I mean the ENV variable

lib/topological_inventory/orchestrator/health_check.rb
roliveri: I think so. It make sense to me.

lib/topological_inventory/orchestrator/health_check.rb
lindgrenj6: Added to `bin/topological_inventory-orchestrator`

lib/topological_inventory/orchestrator/health_check.rb
lindgrenj6: I updated this to mimic the behavior of the default `http` livenessProbe in OpenShift. Basically now it'll register anything that is 400 (or higher) as error. That way if we get a redirect it's still fine.

lib/topological_inventory/orchestrator/health_check.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      private
+
+      def check_url(uri)
+        logger.info("[HealthCheck] Checking URL: #{uri}")
+
+        resp = Net::HTTP.get_response(uri.host, "/health", uri.port)
+        # Checking if the response code is 400 or larger
+        # this is the exact same behavior as OCP's http livenessProbe
+        raise if resp.code.to_i <= 400
```
slemrmartin, line 45: @lindgrenj6 >= 400 :wink:

lindgrenj6: :facepalm: if I had a nickel every time I've made that mistake...

slemrmartin (approved):
:+1: LGTM

roliveri (approved):

PR #119, opened 2020-12-03
Collector OpenShift Services

slemrmartin:
**Issue**: https://github.com/RedHatInsights/topological_inventory-api/issues/320

Deployment of Service objects for Collectors. They're exposing port 9394 for ServiceMonitor (prometheus metrics)

---

[RHCLOUD-10951](https://issues.redhat.com/browse/RHCLOUD-10951)

### DIFF
--- a/lib/topological_inventory/orchestrator/config_map.rb
+++ b/lib/topological_inventory/orchestrator/config_map.rb
@@ -5,14 +5,14 @@ module TopologicalInventory
   module Orchestrator
     # The ConfigMap is identified by label LABEL_COMMON (not unique)
     # And by UUID, which is unique (data.uid)
-    # This UUID is also written to associated deployment_config and secret
+    # This UUID is also written to associated deployment_config, secret and service
     class ConfigMap < OpenshiftObject
       LABEL_COMMON = "tp-inventory/collectors-config-map".freeze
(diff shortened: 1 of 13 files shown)


slemrmartin:
Ready for review, tested on CI (mslemr-test namespace)

lindgrenj6 (commented):
Looks good, just some rubocop issues. Will merge once those are fixed. 

Initially I was worried we didn't have the prometheus annotations on the deployment configs, looks like that is not the case :+1:

slemrmartin:
@lindgrenj6 some of rubocop warnings don't make sense in the context. Test kube_client has to be compatible with real kube_client and named arguments can't have _ prefix. 

Other warnings are made in context of the code around the new one. So I rather chose code similar for the other openshift objects. 
Fixing these rubocop issues will make sense only with fixing other methods not related to this PR. If you think it's necessary, it'll be better to fix in following PR.

lindgrenj6:
@slemrmartin sure, I think we should fix them still (maybe just run `rubocop -a`) and have a new PR for it. That way we don't have these issues every time etc. 

The `_prefix` one won't change though, I agree on that one.

lindgrenj6 (approved):

PR #121, opened 2021-01-05
Cleanup Orphan Collector Pods

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-11145

Just fixing this weird bug that happens. Orphan collectors will just pop up for some reason, no DC/RC just orphans. This cleans them up.

----

Eventually going to re-work orchestrator to set up dependencies of the deployments on the configmaps, e.g. the deployments and secrets will know that the configmap is their "parent" so if we delete the configmap, OpenShift will delete everything for us. That will eliminate a lot of the weird cleanup code in orchestrator.


slemrmartin (approved):
I believe it'll fix the bug https://issues.redhat.com/browse/RHCLOUD-11412

